PR #10, opened 2020-12-18
feat(scanner): Add Unquoted

zmstone:
This is a simplified support for unquoted strings in keys and values


zmstone:
closed in favour of #11

Issue #11, opened 2020-12-18
string concatenation with substitution


zmstone:
e.g.
`a = b${c}`
`a = ${c}"quoted+string"`

PR #12, opened 2020-12-21
feat(src): support variable substitution and concatenation

z8674558:
https://github.com/emqx/hocon/issues/11

In this change, we find `key` at a scanner level. Otherwise the parser would be hard to distinguish `key` from `string`. 
For example
```
a=foo
b=${a}bar"baz"
c=baz
```
the tokens for this HOCON would be
```
string = string string = variable string string string string = string
```
To parse this, we have to tell the parser that `string is key only when = operator follows`, which is hard to describe in yecc (because of shift/reduce conflicts).

By introducing `key` at a scanner level, now the tokens are
```
key string key variable string string string key string
```

### DIFF
--- a/src/hocon_scanner.xrl
+++ b/src/hocon_scanner.xrl
@@ -41,8 +41,8 @@ Ignored             = {WhiteSpace}|{NewLine}|{Comment}
 %% Punctuator
 Punctuator          = [{}\[\],]
 
-%% KeyValue Operator
-KV                  = [:=]
+%% KeyValue
+KvSeparator         = [:=]
 
 %% Bool
 Bool                = true|false|on|off
@@ -53,10 +53,6 @@ Unquoted            = {Letter}[A-Za-z0-9_\.@%\-\|]*
 
 %% Bool
 Bool                = true|false|on|off
-
-%% Name(Atom in Erlang)
-Letter              = [A-Za-z]
-Name                = {Letter}[A-Za-z0-9_\.@-]*
 
 %% Integer
 Digit               = [0-9]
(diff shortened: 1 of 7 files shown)


zmstone:
is concatenated key allowed in the spec?

SGA-kiyofuji:
> is concatenated key allowed in the spec?

Not explicitly described in spec, and I assume it is not allowed. key-value like `a"b"=c` will raise parse error.

This is what they say in the spec:

> a key is a string JSON would have to the left of : and a value is anything JSON would have to the right of :. i.e. the two halves of an object field.

z8674558:
> is concatenated key allowed in the spec?
Not explicitly mentioned, and I assume it is not allowed. key-value like a"b"=c will raise parse error.

> This is what they say in the spec:
a key is a string JSON would have to the left of : and a value is anything JSON would have to the right of :. i.e. the two halves of an object field.

zmstone (commented):
> > is concatenated key allowed in the spec?
> 
> Not explicitly mentioned, and I assume it is not allowed. key-value like a"b"=c will raise parse error.
> 
> This is what they say in the spec:
> 
> > a key is a string JSON would have to the left of : and a value is anything JSON would have to the right of :. i.e. the two halves of an object field.

found this in the spec:
```
Because path expressions work like value concatenations, you can have whitespace in keys:

a b c : 42
is equivalent to:

"a b c" : 42
```
I find it confusing though, and not sure how useful it is.
Maybe we can simply say that we are not going to support it.
If you agree, please describe it in README.

zmstone:
string values are always binary now.
I'd suggest to use binary for key names and for value strings,
including substitution reference names.

src/hocon_scanner.xrl
```diff
@@ -39,7 +39,17 @@ Comment             = (#|//)[^{LineFeed}]*
 Ignored             = {WhiteSpace}|{NewLine}|{Comment}
 
 %% Punctuator
-Punctuator          = [{}\[\]:=,]
+Punctuator          = [{}\[\],]
+
+%% KeyValue Operator
+KV                  = [:=]
```
zmstone, line 45: NIT: maybe rename to `KvSeparactor`

zmstone:
sent a pr to propose some changes: https://github.com/z8674558/hocon/pull/1/files
feel free to either merge (in this case,do not create a merge commit), or just pick the changes.

z8674558:
ty

zmstone (approved):

PR #20, opened 2020-12-24
fix(scanner): Allow float point number to omit integer part when it's 0

zmstone:


z8674558 (approved):
thank you. I don't know if supporting float without integer is helpful for users, though

PR #21, opened 2020-12-24
feat(src): support object merging and recursive variable subst

z8674558:


zmstone:
Please add test cases in `test/hocon_test.erl`.
files in `sample-configs` dir is a collection from  https://github.com/lightbend/config

src/hocon.erl
```diff
@@ -107,7 +107,7 @@ parse(Tokens) ->
 
 -spec(transform(config(), ctx()) -> {ok, config()}).
 transform(Config, Ctx) ->
-    try include(substitute(Config), Ctx) of
+    try include(Config, Ctx) of
```
zmstone, line 110: missing substitute ?

z8674558: substitution is now done by resolve/1

z8674558: which means that substitution occurs before the kv-list is converted into map.

z8674558:
Closing in favor of https://github.com/emqx/hocon/pull/22

PR #26, opened 2021-01-06
run elvis, eunit, gitlint on github actions

z8674558:
example: https://github.com/z8674558/hocon/pull/2

### DIFF
--- a/elvis.config
+++ b/elvis.config
@@ -4,13 +4,10 @@
    [
     {config,
      [#{dirs => ["src"],
-        filter => "hocon.erl",
+        filter => "*.erl",
+        ignore => [hocon_parser, hocon_scanner],
         ruleset => erl_files
        },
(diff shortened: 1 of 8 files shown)


elvis.config
```diff
@@ -0,0 +1,22 @@
+[
+ {
+   elvis,
+   [
+    {config,
+     [#{dirs => ["src"],
+        filter => "hocon.erl",
```
zmstone, line 7: change to `"*.erl"`
and add  `ignore => [hocon_parser, hocon_scanner]` to ignore them.

zmstone:
```
      #{dirs => ["src"],
        filter => "*.erl",
        ruleset => erl_files,
        ignore => [hocon_parser, hocon_scanner]
      },
```

z8674558: ty for the hint. fixed

PR #29, opened 2021-01-07
fix #25 extract occurs resolve and just returns kvlist

### LINKED ISSUE #25: substitution lookup error
file a
```
foo.x: 45
y: 2
```
it can be parsed as expected: `#{foo => #{x => 45},y => 2}`

file b
```
foo={x: 45}
y: 2
```
it is also parsed as expected: `#{foo => #{x => 45},y => 2}`

however, if we change the second line to `y: ${foo}`, file a results in exception `{unresolved,[foo]}` (but not file b)

z8674558:


src/hocon.erl
```diff
@@ -171,6 +172,28 @@ do_abspath(Var, ['$root']) ->
...
+do_expand([{Key, {concat, C}}|More], Acc) ->
+    do_expand(More, [create_nested(Key, {concat, do_expand(C, [])})|Acc]);
+do_expand([{Key, Value}|More], Acc) ->
+    do_expand(More, [create_nested(Key, Value)|Acc]);
+do_expand([{Object}|More], Acc) when is_list(Object) ->
+    do_expand(More, [{do_expand(Object, [])}|Acc]);
+do_expand([Other|More], Acc) ->
+    do_expand(More, [Other|Acc]).
+
+create_nested(Key, Value) when is_atom(Key) ->
```
zmstone, line 190: IMO it's better for this function to have its own name instead of sharing with the other two clauses which are self-looping

src/hocon.erl
```diff
@@ -171,6 +172,28 @@ do_abspath(Var, ['$root']) ->
...
+    {concat, [{[Res]}]} = create_nested(paths(Key), Value),
+    Res;
+create_nested([], Value) ->
+    Value;
+create_nested([Path|More], Value) ->
+    {concat, [{[{Path, create_nested(More, Value)}]}]}.
```
zmstone, line 196: I am not quite following why it has to be a `concat`.

z8674558: As all objects are wrapped by `concat` in parser, we have to mock that to make it readable in the following pipeline.

src/hocon.erl
z8674558: advice applied

zmstone (approved):

PR #30, opened 2021-01-08
feat(src): support maybevar and delete unresolved ones in remove_nothing

z8674558:


zmstone:
```
x=${?y}
z=${?x}
```
this results in`#{z=>nothing}`, expected ?

PR #32, opened 2021-01-12
support proplists as output format, stop converting on|off to true|false

z8674558:


src/hocon.erl
```diff
@@ -28,13 +28,44 @@
...
 
+proplists(Map) when is_map(Map) ->
+    proplists(maps:iterator(Map), [], []).
+proplists(Iter, Path, Acc) ->
+    case maps:next(Iter) of
+        {K, M, I} when is_map(M) ->
+            Child = proplists(maps:iterator(M), [atom_to_list(K)| Path], []),
+            proplists(I, Path, lists:append(Child, Acc));
+        {K, [Bin|_More]=L, I} when is_binary(Bin) ->
+            NewList = [binary_to_list(B) || B <- L],
```
zmstone, line 39: it could potentially be utf8 here.
`binary:characters_to_list/2`

src/hocon.erl
```diff
@@ -28,13 +28,44 @@
...
+            ReversedPath = lists:reverse([atom_to_list(K)| Path]),
+            proplists(I, Path, [{ReversedPath, NewList}| Acc]);
+        {K, Bin, I} when is_binary(Bin) ->
+            ReversedPath = lists:reverse([atom_to_list(K)| Path]),
```
zmstone, line 43: not sure we can actually use unicode for keys, maybe worth testing.
and all the binary atom conversions should take utf8 in to consideration.

src/hocon.erl
```diff
@@ -28,13 +28,44 @@
...
+            proplists(I, Path, [{ReversedPath, V}| Acc]);
+        none ->
+            Acc
+    end.
+
 -spec(load(file:filename()) -> {ok, config()} | {error, term()}).
 load(Filename0) ->
+    load(Filename0, [{format, map}]).
+
+-spec(load(file:filename(), list()) -> {ok, config()} | {error, term()}).
```
zmstone, line 56: please add a type spec for `Opts`, and I'd go with map.
e.g.
```
#{ format => map | proplist
 , key => string | atom
 , string => list | binary
 }
```

zmstone (approved):

PR #35, opened 2021-01-13
move converting functions to hocon.erl & run dialyzer on github actions

z8674558:


src/hocon.erl
```diff
@@ -50,6 +60,19 @@ proplists(Iter, Path, Acc) ->
             Acc
     end.
 
+convert_value(L, Map) when is_list(L) ->
+    lists:foldl(fun (E, A) -> do_convert_value(E, A) end, Map, L).
```
zmstone, line 64: better to create a high-order function which wraps around all the converts and call `nested_update_with` only once.

this is because the `Map` to be updated can be large and each iteration will result in a new map then end up to be garbage collected.

src/hocon.erl
```diff
@@ -22,10 +22,20 @@
 
 -type config() :: map().
 -type ctx() :: map().
--type opts() :: map().
+-type opts() :: #{format => map | proplists,
+                  convert => list()}.
```
zmstone, line 26: make it a `[convert()]`
and then
```
convert() :: duration | bytesize | percent | onoff | convert_func()
convert_func() :: fun((term()) -> term()).
```

src/hocon.erl
```diff
@@ -434,6 +466,21 @@ nested_put([Key], Val, Map) ->
 nested_put([Key|Paths], Val, Map) ->
     merge(Key, nested_put(Paths, Val, #{}), Map).
 
+nested_update_with(Fun, Map) ->
+    nested_update_with(Fun, Map, maps:iterator(Map)).
+
+nested_update_with(Fun, Map, I) ->
```
zmstone, line 472: we may want to support converting Key from `atom()` to `binary()` or `string()`,
so it's better to create another Map instead of updating in-place. ?

src/hocon.erl
```diff
@@ -434,6 +466,21 @@ nested_put([Key], Val, Map) ->
...
+    case maps:next(I) of
+        {Key, ChildMap, Next} when is_map(ChildMap) ->
+            Updated = #{Key => nested_update_with(Fun, ChildMap)},
+            nested_update_with(Fun, do_deep_merge(Map, Updated), Next);
```
zmstone, line 476: why deep merge? key is unique and value is already updated recursively

src/hocon.erl
z8674558: thank you!!!

src/hocon.erl
z8674558: nice idea, created `pipeline_fun/1` to handle it.

src/hocon.erl
z8674558: fixed

src/hocon.erl
z8674558: true

src/hocon.erl
```diff
@@ -50,6 +62,35 @@ proplists(Iter, Path, Acc) ->
...
+convert_value(ConvertFunctions, Map) when is_list(ConvertFunctions) ->
+    Resolved = resolve_convert_fun(ConvertFunctions),
+    do_convert_value(pipeline_fun(Resolved), Map).
+
+do_convert_value(Fun, Map) ->
+    do_convert_value(Fun, maps:iterator(Map), #{}).
+do_convert_value(Fun, I, NewMap) ->
+    case maps:next(I) of
+        {K, M, Next} when is_map(M) ->
+            do_convert_value(Fun, Next, maps:merge(NewMap, #{K => do_convert_value(Fun, M)}));
```
zmstone, line 74: nit: `NewMap#{K => do_convert_value(Fun, M)}` should do.

zmstone (approved):

test/hocon_tests.erl
```diff
@@ -186,9 +186,20 @@ cuttlefish_proplists_test() ->
                        {["cluster", "name"], "emqxcl"},
                        {["cluster", "discovery"], "manual"},
                        {["cluster", "autoheal"], "on"},
-                       {["cluster", "autoclean"], 300000}]},
+                       {["cluster", "autoclean"], "5m"}]},
                  hocon:load("etc/node.conf", #{format => proplists})).
 
+apply_opts_test() ->
```
zmstone, line 192: maybe add a test to cover caller supplied converter function ?

PR #37, opened 2021-01-15
chore(test): exclude unsupported files from test

z8674558:


src/hocon.erl
```diff
@@ -515,8 +516,8 @@ pipeline(Input, Ctx, [Fun | Steps]) ->
     pipeline(Output, Ctx, Steps);
 pipeline(Result, _Ctx, []) -> Result.
 
-scan_error(Line, ErrorInfo) ->
-    throw({scan_error, format_error(Line, ErrorInfo)}).
+scan_error(Line, ErrorInfo, Ctx) ->
+    throw({scan_error, format_error(Line, ErrorInfo, Ctx)}).
 
 parse_error(Line, ErrorInfo) ->
```
zmstone, line 522: same here ?

sample-configs/subdir/foo.conf
```diff
@@ -1,5 +1,5 @@
 foo=42
 # included without file()
 include "bar.conf"
-# included using file()
+# included using file() #### So far no file include support
```
zmstone, line 4: this should not have been commented out (my bad)
since we are expecting an error in test, maybe we can revert to the original content.

src/hocon.erl
z8674558: fixed

sample-configs/subdir/foo.conf
z8674558: `# included using file()` is a comment. `include file("bar-file.conf")` on line 5 is not commented out.

zmstone (approved):

PR #41, opened 2021-01-20
fix(src): forbid whitespace between unquoted

z8674558:


zmstone:
For future reference:

Our implementation (before this PR) is, to concatenate unquoted strings without preserving whitespaces.
This behaviour is different from the spec, but consistent and concise,
if people wanted whitespace, they can always quote.

The motivation of this change originated from some corner cases where `foo include require`
(i.e. an incomplete include expression), got concatenated into `fooincluderequire`

After reviewing the changes, we realise that forbidding spaces in between will only increase
confusion of the concatenation logic.
So we decided to close it.

Issue #42, opened 2021-01-28
substitution fails to get a merged object


z8674558:
```
a={x:1}
a={y:2}
b=${a}
```

returns `#{a => #{x => 1,y => 2},b => #{y => 2}}` (expected `b => #{x => 1,y => 2}`)

zmstone:
fixed  in #43

PR #43, opened 2021-01-28
fix(src): merge old ResolvedValue when possible

z8674558:
And add interop test, which reminded me of this bug.
https://github.com/emqx/hocon/issues/42


zmstone:
> Within an field value or array element, if only non-newline whitespace separates the end of a first array or object or substitution from the start of a second array or object or substitution, the two values are concatenated. Newlines may occur within the array or object, but not between them. Newlines between prevent concatenation.

It seems that the spec does not allow array concatenation to span multiple lines.
Maybe something worth mentioning in README.md

zmstone (approved):

zmstone (changes requested):
sorry, approved too fast, found an edge case:
```
a=[{k1=1}] [{k2=2}]
a=[{k3=1}] [{k2=2}]
b=${a}
```
is loaded as:
```
#{a => [#{k3 => 1},#{k2 => 2}],
  b => [#{k1 => 1},#{k2 => 2},#{k3 => 1},#{k2 => 2}]}.
```
but it should be :
```
#{a => [#{k3 => 1},#{k2 => 2}],
  b => [#{k3 => 1},#{k2 => 2}]}.
```

sample-configs/json/test01.conf.json
```diff
@@ -0,0 +1 @@
...":"1 milli","minusLargeNanos":"-4878955355435272204ns","plusLargeNanos":"+4878955355435272204ns","second":"1s","secondAsNumber":1000,"secondsList":["1s","2seconds","3 s",4000]},"floats":{"fortyTwoPointOne":42.1,"fortyTwoPointOneAgain":42.1,"pointThirtyThree":".33","pointThirtyThreeAgain":".33"},"ints":{"fortyTwo":42,"fortyTwoAgain":42},"memsizes":{"halfMeg":"0.5M","meg":"1M","megAsNumber":1048576,"megsList":["1M","1024K",1048576],"yottabyte":"1YB","yottabyteList":["1YB","0.5YB"]},"nulls":{"null":null,"nullAgain":null},"periods":{"day":"1d","dayAsNumber":2,"month":"5 mo","week":"3 weeks","year":"8y"},"strings":{"a":"a","abcd":"abcd","abcdAgain":"abcd","b":"b","c":"c","concatenated":"null bar 42 baz true 3.14 hi","d":"d","double":"3.14","doubleStartingWithDot":".33","false":"false","no":"no","null":"null","number":"57","true":"true","yes":"yes"},"system":{"concatenated":"Your Java version is  and your user.home is ","home":"/Users/yudai","path":"/Users/yudai/.conscript/bin:/Users/yudai/rebar3/bin:/Users/yudai/.asdf/shims:/usr/local/opt/asdf/bin:/Users/yudai/.nodebrew/current/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin","pwd":"/Users/yudai/IdeaProjects/hocon","shell":"/bin/zsh"}}
```
zmstone, line 1: Users/yudai ...

test/hocon_tests.erl
```diff
@@ -57,20 +60,52 @@ test_file_load("test10", F) ->
...
+test_interop("test08", _, _) ->
+    ok;
+test_interop("test10", _, _) ->
+    ok;
+test_interop(_, Conf, Json) ->
+    case hocon:load(Conf) of
+        {error, _} ->
+            ok;
+        Res ->
+            ?assertEqual(Res, hocon:load(Json))
```
zmstone, line 106:
```suggestion
            ?assertEqual(hocon:load(Json), Res)
```

src/hocon.erl
```diff
@@ -306,6 +306,27 @@ lookup(Var, [_Other|More], ResolvedValue) ->
...
+typeof({concat, [X|_]}) when is_list(X) ->
+    array;
+typeof({concat, [{X}|_]}) when is_list(X) ->
+    object;
+typeof({concat, _Other}) ->
+    string.
+
+maybe_merge({concat, Old}, {concat, New}) ->
+    case {typeof({concat, Old}), typeof({concat, New})} of
+        {array, array} ->
```
zmstone, line 320: since we are trying to look up a field value here,
the field value can only be concatenated when it's an object, (or object of objects),
but can never be concatenated when it's an array, (or array of arrays).

```
is_object([{concat, X}|_]) -> is_object(X);
is_object([{X}|_]) -> is_list(X);
is_object(_) -> false.
```
Then only append `Old` and `New` when they are both objects.

sample-configs/json/test01.conf.json
z8674558: It is ignored in the test anyway, so I've just excluded this from being generated.

src/hocon.erl
```diff
@@ -306,6 +306,27 @@ lookup(Var, [_Other|More], ResolvedValue) ->
...
+    typeof({concat, X});
+typeof({concat, [X|_]}) when is_list(X) ->
+    array;
+typeof({concat, [{X}|_]}) when is_list(X) ->
+    object;
+typeof({concat, _Other}) ->
+    string.
+
+maybe_merge({concat, Old}, {concat, New}) ->
+    case {typeof({concat, Old}), typeof({concat, New})} of
```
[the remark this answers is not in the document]
z8674558: nice idea

zmstone (approved):

PR #45, opened 2021-02-01
fix(concat): check if all elements have same type

z8674558:


src/hocon.erl
```diff
@@ -400,11 +400,29 @@ do_concat(Concat) ->
 do_concat([], [{}]) ->
     {[]};
 do_concat([], [Object| _Objects] = Acc) when is_tuple(Object) ->
```
zmstone, line 402: how about 
```
-define(IS_OBJECT(O), (is_tuple(O) andalso size(O) =:= 1)).
```

and below to replace `{[...]}`.
```
-define(OBJECT(FIELDS), {FIELDS}).
```

src/hocon.erl
```diff
@@ -400,11 +400,29 @@ do_concat(Concat) ->
...
-    {lists:reverse(Acc)};
+    case lists:all(fun is_tuple/1, Acc) of
+        true ->
+            {lists:reverse(Acc)};
+        false ->
+            % TODO: print line number
```
zmstone, line 407: line number info has been dropped at this point, right ?

src/hocon_token.erl
```diff
@@ -201,9 +201,14 @@ scan_error(Line, ErrorInfo, Ctx) ->
...
 
-format_error(Line, ErrorInfo, Ctx) ->
+format_error(Line, ErrorInfo, #{filename := F}) ->
     binary_to_list(
         iolist_to_binary(
             [ErrorInfo,
              io_lib:format(" in line ~w. file: ~p",
-                           [Line, hd(hocon_util:get_stack(filename, Ctx))])])).
+                           [Line, hd(F)])]));
+format_error(Line, ErrorInfo, _Ctx) ->
```
zmstone, line 210: is this to make the `binary` API work ?
maybe it's easier if `hocon:binary/1` creates a `'nofile'` to make it easier.

src/hocon.erl
```diff
@@ -400,11 +400,29 @@ do_concat(Concat) ->
...
+            throw({concat_error, Acc})
```
zmstone, line 408: maybe it's easier to reason about if acc is reversed here.

src/hocon.erl
```diff
@@ -400,11 +400,29 @@ do_concat(Concat) ->
...
+    end;
 do_concat([], [String| _Strings] = Acc) when is_binary(String) ->
-    iolist_to_binary(lists:reverse(Acc));
+    case lists:all(fun is_binary/1, Acc) of
+        true ->
+            iolist_to_binary(lists:reverse(Acc));
+        false ->
+            throw({concat_error, Acc})
```
zmstone, line 415: maybe it's easier to reason about if acc is reversed here.

src/hocon.erl
```diff
@@ -400,11 +400,29 @@ do_concat(Concat) ->
...
+    end;
 do_concat([], [Array| _Arrays] = Acc) when is_list(Array) ->
-    lists:append(lists:reverse(Acc));
+    case lists:all(fun is_list/1, Acc) of
+        true ->
+            lists:append(lists:reverse(Acc));
+        false ->
+            throw({concat_error, Acc})
```
zmstone, line 422: maybe it's easier to reason about if acc is reversed here.

src/hocon.erl
z8674558:
```
-define(IS_OBJECT(O), (is_tuple(O) andalso size(O) =:= 1 andalso is_list(element(1, O)))).
-define(IS_FIELD(F), (is_tuple(F) andalso size(F) =:= 2)).
-define(OBJECT(FIELDS), {FIELDS}).
-define(FIELDS(OBJECT), element(1, OBJECT)).
```

src/hocon.erl
z8674558: No, line number info is discarded in parser, when we use `value_of(Token) -> element(3, Token).`.

src/hocon_token.erl
z8674558: fix

src/hocon.erl
zmstone: right, in that case, we can raise the error with the line number for the first element ?
i.e. no to print it.

z8674558: We can do it after we make the parser preserve line number.

zmstone (approved):

PR #46, opened 2021-02-04
feat(src): preserve line number after parse

z8674558:


zmstone (commented):
Since we are enriching keys with line number and values with type information.
I'm thinking maybe we can take one step further: box all the values, not only the primitives (terminals), but also the complex values: objects and arrays.

The type tree expressed as Eralng type specs:
```
-type boxed() :: #{type := hocon_type(), value := inbox(), location => location()}.
-type inbox() :: primitive() | [{Key :: atom(), boxed()}] | [boxed()].
-type primitive() :: number() | atom() | string() | binary().
-type hocon_type() :: null | bool | integer | float | string | array | object | hocon_intermediate_type(). 
-type hocon_intermediate_type() :: variable | include | concat. %% to be reduced
-type location() :: #{filename(), line_number()}.
```

z8674558:
closing in favor of https://github.com/emqx/hocon/pull/47

PR #47, opened 2021-02-08
feat(src): make tokens have line number and filename, and print them when error

z8674558:


src/hocon_token.erl
```diff
@@ -17,8 +17,22 @@
...
 
 -export([read/1, scan/2, trans_key/1, parse/2, include/2]).
+-export([value_of/1]).
 
--type include() :: #{filename => binary(), required => boolean()}.
+-export_type([boxed/0, inbox/0]).
+
+-type boxed() :: #{type := hocon_type(),
+                   value := inbox(),
+                   line => string(),
```
zmstone, line 26: why is line a string ?

test/hocon_tests.erl
```diff
@@ -267,19 +269,70 @@ merge_when_resolve_test() ->
                         b => [#{k3 => 1}, #{k2 => 2}]}},
                  hocon:binary("a=[{k1=1}] [{k2=2}],a=[{k3=1}] [{k2=2}],b=${a}")).
 
-concat_error_test_() ->
-    [ ?_assertEqual({error, {concat_error, [{x, 1}, [1, 2]]}},
+concat_error_binary_test_() ->
+    [ ?_assertEqual({error,
+                     {concat_error,
+                      <<"failed to concat [{x,1},[1,2]] in file undefined at line 1">>}},
```
zmstone, line 275: to be search friendly, we should use `key_words_for_error_message`,
e.g. `failed_to_concat [{x,1},[1,2]] in_file <filename> at_line <linenumber>"`

and `in_file <filename>` can be skipped if it's `undefined`

src/hocon_postprocess.erl
```diff
@@ -0,0 +1,180 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2020 EMQ Technologies Co., Ltd. All Rights Reserved.
```
zmstone, line 2: 2021

src/hocon_token.erl
z8674558: fixed

src/hocon_parser.yrl
```diff
@@ -17,23 +17,23 @@ Terminals
   include key required.
 
 Rootsymbol hocon.
-hocon -> '{' fields endobj : make_object('$2').
-hocon -> fields : make_object('$1').
+hocon -> '{' fields endobj : make_object(0, '$2').
+hocon -> fields : make_object(0, '$1').
```
z8674558, line 21: Now root object has a line number 0 and `make_object` becomes more simple.

However, I am not sure if it makes sense to `try to take the first field’s line number for the object’s line number if 0 is passed in` as you've said. Because I do not imagine any usecases that would need a line number for root object. Obviously a root object does not cause `resolve_error`, and when it comes to `concat_error`, sometimes a root object can be concatenated like this:
```
a = {include "foo.conf}
b = "xyz"${a}
```
but in this case, we want to print the location for `b`, which is `line 2`. 

This is why I propose we leave it to `0`.

test/hocon_tests.erl
z8674558: changes applied

src/hocon_postprocess.erl
zmstone: still 2020

test/hocon_tests.erl
```diff
@@ -267,19 +269,70 @@ merge_when_resolve_test() ->
                         b => [#{k3 => 1}, #{k2 => 2}]}},
                  hocon:binary("a=[{k1=1}] [{k2=2}],a=[{k3=1}] [{k2=2}],b=${a}")).
 
-concat_error_test_() ->
-    [ ?_assertEqual({error, {concat_error, [{x, 1}, [1, 2]]}},
+concat_error_binary_test_() ->
+    [ ?_assertEqual({error,
+                     {concat_error,
+                      <<"failed to concat [{x,1},[1,2]] at_line 1">>}},
```
zmstone, line 275: nit: `faied_to_concat` ?

zmstone (dismissed):

test/hocon_tests.erl
z8674558: also `failed to resolve` is fixed to `failed_to_resolve`

zmstone (approved):

PR #50, opened 2021-02-23
test(elvis): spaces around |, no spaces after ( and before )

z8674558:


z8674558:
@zmstone trying out new rules here. Once you approve, I will apply this change on emqx.git.

zmstone:
can we add `|` and `||` here:
https://github.com/inaka/elvis_core/blob/c4ffeb071ec7c01e5229091f32ce2e091f2cfbcf/src/elvis_style.erl#L323

maybe even a PR to upstream

zmstone:
these are the defaults.
https://github.com/inaka/elvis_core/blob/c4ffeb071ec7c01e5229091f32ce2e091f2cfbcf/src/elvis_style.erl#L145-L150

 maybe we should enable all in our elvis.config

z8674558:
https://github.com/inaka/katana-code/blob/7e635609446b6f1e4cdd233f6dcb1d5ab34d99b8/src/ktn_code.erl#L25
This is the all types of tokens we can get from an `.erl` file.

z8674558:
Some of operators are packed to `op` map. `|` and `||` are not.

https://github.com/inaka/katana-code/blob/7e635609446b6f1e4cdd233f6dcb1d5ab34d99b8/src/ktn_code.erl#L627

scripts/elvis_more_rules.erl
```diff
@@ -0,0 +1,155 @@
...
+    throw({invalid_file, File}).
+
+find_encoding(Content) ->
+    case epp:read_encoding_from_binary(Content) of
+        none -> utf8;
+        Enc  -> Enc
+    end.
+
+
+% eunit
```
k32, line 99: Perhaps wrap tests in `-ifdef(TEST).`?

z8674558: ty, I was not aware of it.

zmstone:
> Some of operators are packed to `op` map. `|` and `||` are not.
> 
> https://github.com/inaka/katana-code/blob/7e635609446b6f1e4cdd233f6dcb1d5ab34d99b8/src/ktn_code.erl#L627

I believe `|` is `cons`.
https://github.com/inaka/katana-code/blob/7e635609446b6f1e4cdd233f6dcb1d5ab34d99b8/src/ktn_code.erl#L36

zmstone:
Maybe move this module to our elvis fork ?
so we can reuse without having to copy the module to another project ?
https://github.com/emqx/elvis

z8674558:
> Maybe move this module to our elvis fork ?

sure. by doing so we should be write this more simple.

zmstone:
this PR can be reopened, we still need the config changes and HOCON code changes ?

PR #53, opened 2021-04-06
print special message when unicode / support week, fortnight

z8674558:


src/hocon_postprocess.erl
```diff
@@ -187,6 +187,8 @@ do_duration(Str, Sum) ->
             end
     end.
 
+calc_duration(Val, "f")  -> Val * ?FORTNIGHT;
```
z8674558, line 190: this is just to pass the tests in cuttlefish
https://github.com/emqx/cuttlefish/blob/83887e3ef71d65ec6a066726c97b2bbce87f7800/src/cuttlefish_duration.erl#L138

zmstone:
We may need to support unicode. 
Call `unicode:characters_to_binary`
and `characters_to_list` to convert between string and binary.

zmstone (approved):

src/hocon_postprocess.erl
```diff
@@ -170,23 +170,48 @@ duration(Other) ->
...
+                    case get_decimal(Val) of
+                        D ->
+                            Sum + calc_duration(D, string:lowercase(Unit))
+                    end;
                 nomatch ->
                     skip
             end
     end.
 
+get_decimal([$. | _]=Num) ->
```
z8674558, line 198: another chore to make test pass
https://github.com/emqx/cuttlefish/blob/83887e3ef71d65ec6a066726c97b2bbce87f7800/src/cuttlefish_duration.erl#L100

src/hocon_postprocess.erl
```diff
@@ -170,23 +170,50 @@ duration(Other) ->
...
+    get_decimal(["0" | Num]);
+get_decimal(Num) ->
+    case string:to_integer(Num) of
```
zmstone, line 201: why not `string:to_float/1` ?

src/hocon_postprocess.erl
```diff
@@ -170,23 +170,50 @@ duration(Other) ->
...
+                    skip
+            end;
         nomatch ->
-            {ok, LastMP} = re:compile("^([0-9]+)(d|D|h|H|m|M|s|S|ms|MS)$"),
+            {ok, LastMP} = re:compile("^([0-9\.]+)(f|F|w|W|d|D|h|H|m|M|s|S|ms|MS)$"),
             case re_run_first(Str, LastMP) of
                 {match, [Val, Unit]} ->
-                    Sum + calc_duration(list_to_integer(Val), string:lowercase(Unit));
+                    case get_decimal(Val) of
+                        D ->
```
zmstone, line 190: single clause `case` expression?
maybe missed a `{error, Reason}` clause ?

src/hocon_postprocess.erl
```diff
@@ -170,23 +170,50 @@ duration(Other) ->
...
+            I;
+        {Int, [$. | MaybeDecimal]} ->
+            case string:to_integer(MaybeDecimal) of
+                {Decimal, []} ->
+                    Int + Decimal / math:pow(10, length(MaybeDecimal));
+                _ ->
+                    {error, badarg}
+            end;
+        _ ->
+            error
```
zmstone, line 212: it's better if `get_decimal` returns either `error` or `{error, Reason}`, but not a mix

src/hocon_postprocess.erl
zmstone: maybe this:
```
get_decimal(Num) ->
    case string:to_float(Num) of
        {F, []} -> F;
        {error, no_float} -> {I, []} = string:to_integer(Num), I
    end
```

zmstone: given that the input `Num` was matched by regex pattern `([0-9\.]+)`, it's impossible to have other outputs from `string:to_float` or `string:to_integer`, so we can use assertive matches.
otherwise the lines will never be covered in tests (unless stub the inputs).

src/hocon_postprocess.erl
zmstone: `D = get_decimal(Val)`

src/hocon_postprocess.erl
```diff
@@ -170,23 +170,41 @@ duration(Other) ->
...
 do_duration(Str, Sum) ->
-    {ok, MP} = re:compile("^([0-9]+)(d|D|h|H|m|M|s|S|ms|MS)([0-9]+.+)"),
+    {ok, MP} = re:compile("^([0-9\.]+)(f|F|w|W|d|D|h|H|m|M|s|S|ms|MS)([0-9\.]+.+)"),
     case re_run_first(Str, MP) of
         {match, [Val, Unit, Next]} ->
-            do_duration(Next, Sum + calc_duration(list_to_integer(Val), string:lowercase(Unit)));
+            case get_decimal(Val) of
+                D when is_integer(D) orelse is_float(D) ->
+                    do_duration(Next, Sum + calc_duration(D, string:lowercase(Unit)));
+                {error, badarg} ->
```
zmstone, line 182: this is not possible, is it ?

z8674558: ty

zmstone (approved):

PR #56, opened 2021-04-22
feat(hocon_schema): add schema engine (with typerefl)

z8674558:


sample-schemas/demo_schema.erl
```diff
@@ -0,0 +1,66 @@
...
+
+keys() ->
+    [ {"foo.setting", fun foo__setting/1}
+    , {"foo.endpoint", fun foo__endpoint/1}
+    ].
+
+foo__setting(mapping) -> "app_foo.setting";
+foo__setting(type) -> string();
+foo__setting(converter) -> converter(string);
+foo__setting(validator) -> type_checker(foo__setting(type));
```
k32, line 21: I think it always makes sense to validate the type. (In order to omit this check, one can use `any()` type). Therefore, call to `typerefl:type_check` function should be made by the "framework" code, and `validator` clause should be optional, and reserved for custom validators. 
This will reduce the amount of boilerplate in the schema.

sample-schemas/demo_schema.erl
```diff
@@ -0,0 +1,66 @@
...
+-type ip() :: ipv4() | ipv6().
+-type ipv4() :: {0..255, 0..255, 0..255, 0..255}.
+-type ipv6() :: {0..65535, 0..65535, 0..65535, 0..65535, 0..65535, 0..65535, 0..65535, 0..65535}.
+
+-reflect_type([ip/0]).
+
+-export([keys/0]).
+
+keys() ->
+    [ {"foo.setting", fun foo__setting/1}
```
zmstone, line 14: we do not need to follow `__` for `.` naming convention here
that was for the old proposal which we try to get field names from exported functions list.

sample-schemas/demo_schema.erl
zmstone: agree to @k32 .
if validator clause returns a `fun`, it applies on top of typerefl check result.

src/hocon_schema.erl
```diff
@@ -0,0 +1,106 @@
...
+
+map(Schema, RichMap) ->
+    Keys = apply(Schema, keys, []),
+    do_map(Keys, RichMap, []).
+
+do_map([], _RichMap, Acc) ->
+    Acc;
+do_map([{Key, SchemaFun} | More], RichMap, Acc) ->
+    Value = deep_get(Key, RichMap, value),
+    Converter = SchemaFun(converter),
```
zmstone, line 33: converter could return `undefined`.
we may need a `apply_converter` function.
e.g.
apply_converter(SchemaFun, converter, Value) ->
    case SchemaFun(converter) of
        undefined ->  Value;
        F -> F(Value)
    end.
```

Comment (zmstone) [reply src/hocon_schema.erl:33]: and it's the same for other kinds too if they are optional.

File: src/hocon_schema.erl [hunk 4]
```diff
@@ -0,0 +1,128 @@
...
+
+add_default_validator(undefined, Type) ->
+    add_default_validator([], Type);
+add_default_validator(Validators, Type) ->
+    TypeChecker = fun (Value) ->
+        case typerefl:typecheck(Type, Value) of
+            ok ->
+                true;
+            {error, _} ->
+                false
```
zmstone, line 62: we may need to return informative errors from validator.

src/hocon_schema.erl
```diff
@@ -0,0 +1,128 @@
...
+        end
+    end,
+    [TypeChecker | Validators].
+
+validate(Value, []) ->
```
zmstone, line 67: we can return a list of errors here

src/hocon_schema.erl
```diff
@@ -0,0 +1,128 @@
...
+    Value0 = apply_converter(SchemaFun, Value),
+    Validators = add_default_validator(SchemaFun(validator), SchemaFun(type)),
+    Mapping = string:tokens(SchemaFun(mapping), "."),
+    case {Value0, SchemaFun(default)} of
+        {undefined, undefined} ->
+            do_map(More, RichMap, Acc);
+        {undefined, Default} ->
+            do_map(More, RichMap, [{Mapping, Default} | Acc]);
+        {Value0, _} ->
+            Value1 = validate(Value0, Validators),
```
zmstone, line 42: handle error(s) returned from `validate/2`
maybe accumulate errors in a list (with both original value and converted values)

src/hocon_schema.erl
```diff
@@ -0,0 +1,128 @@
...
+            do_map(More, RichMap, [{Mapping, Value1} | Acc])
+    end.
+
+apply_converter(SchemaFun, Value) ->
+    case SchemaFun(converter) of
+        undefined ->
+            Value;
+        Converter ->
+            Converter(Value)
```
zmstone, line 51: catch converter exception, return as error.

sample-schemas/demo_schema.erl
```diff
@@ -0,0 +1,60 @@
...
+convert_to_ip(Str) when is_list(Str) ->
+    case inet:parse_address(Str) of
+        {ok, Addr} ->
+            Addr;
+        {error, einval} ->
+            Str
+    end;
+convert_to_ip(V) -> V.
+
+convert_to_string(Int) when is_integer(Int) ->
```
k32, line 48: `lists:flatten(io_lib:format("~p", [V]))` ?

sample-schemas/demo_schema.erl
```diff
@@ -0,0 +1,60 @@
...
+
+keys() ->
+    [ {"foo.setting", fun setting/1}
+    , {"foo.endpoint", fun endpoint/1}
+    , {"foo.greet", fun greet/1}
+    ].
+
+setting(mapping) -> "app_foo.setting";
+setting(type) -> string();
+setting(converter) -> converter(string);
```
k32, line 21: `converter` seems like an ambiguous term to me. Change to `to_string`?

zmstone: or `fun to_string/1`

sample-schemas/demo_schema.erl
```diff
@@ -0,0 +1,60 @@
...
+
+-type ip() :: ipv4() | ipv6().
+-type ipv4() :: {0..255, 0..255, 0..255, 0..255}.
+-type ipv6() :: {0..65535, 0..65535, 0..65535, 0..65535, 0..65535, 0..65535, 0..65535, 0..65535}.
+
+-reflect_type([ip/0]).
+
+-export([keys/0]).
+
+keys() ->
```
zmstone, line 13: maybe rename it to:
`fields() ->`
and add another function
```
namespace(source) -> "foo".
namespace(mapping) -> app_foo.
```
so we do not have to repeat `"foo." and `"app_foo."`

sample-schemas/demo_schema.erl
```diff
@@ -0,0 +1,60 @@
...
+
+-include_lib("typerefl/include/types.hrl").
+
+-type ip() :: ipv4() | ipv6().
+-type ipv4() :: {0..255, 0..255, 0..255, 0..255}.
+-type ipv6() :: {0..65535, 0..65535, 0..65535, 0..65535, 0..65535, 0..65535, 0..65535, 0..65535}.
+
+-reflect_type([ip/0]).
+
+-export([keys/0]).
```
zmstone, line 11: we may need the ability to define multiple scehmas per module
e.g. for nested sub ojbect.
```
-export([names/0, mappings/0, fields/1]).

names() -> [foo, nested].

mappings(foo) -> app_foo;
mappings(_) -> undefined.

fields(foo) -> [{"setting", fun setting/1}, 
{"nested", fun nested/1}
];
fields(nested) ->
[ {"abc", fun abc/1}
].

```

sample-schemas/demo_schema.erl
```diff
@@ -0,0 +1,60 @@
...
+
+keys() ->
+    [ {"foo.setting", fun setting/1}
+    , {"foo.endpoint", fun endpoint/1}
+    , {"foo.greet", fun greet/1}
+    ].
```
zmstone, line 17: we may need to support some built in types like:

`{"foo.flag", bool}` which get translated into `fun hocon_schema_builtin:bool/1`.
same for ip type

sample-schemas/demo_schema.erl
zmstone: with mapping defined at 'object' level, there is no need to define mappings for each field.

sample-schemas/demo_schema.erl
z8674558: since we catch exception in `apply_converter`, now it is ok to be just `integer_to_list`

sample-schemas/demo_schema.erl
z8674558: Sometimes mapping is not one-to-one.

sample-schemas/demo_schema.erl
z8674558: `hocon_schema_builtin:bool` would need to return a mapping name, which is impossible.

z8674558: instead we can omit `converter`. If `converter` is undefined, hocon_schema will look for default converter based on its type

sample-schemas/demo_schema.erl
```diff
@@ -0,0 +1,57 @@
...
+
+setting(mapping) -> "app_foo.setting";
+setting(type) -> string();
+setting(converter) -> fun to_string/1;
+setting(_) -> undefined.
+
+endpoint(mapping) -> "app_foo.endpoint";
+endpoint(type) -> ip();
+endpoint(converter) -> fun to_ip/1;
```
k32, line 26: I think typerefl should support custom `(from|to)_string` callbacks for the types (so all IP addresses are parsed/printed in the same way). What do you think, @z8674558 @zmstone ?

z8674558: sounds nice. however, type conversion is not always `(from|to)_string`? e.g. when we deal with `percent` type, we would like to use `(from|to)_float`.

sample-schemas/demo_schema.erl
```diff
@@ -0,0 +1,57 @@
...
+endpoint(_) -> undefined.
+
+greet(mapping) -> "app_foo.greet";
+greet(type) -> binary();
+greet(validator) ->
+    fun (Value) ->
+        case Value =:= "hello" of
```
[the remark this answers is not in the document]
z8674558: ty!!!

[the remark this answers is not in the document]
k32: Should be available in version 0.2.0

sample-schemas/demo_schema.erl
k32: Could you give an example of how conversion for percent type should work?

k32: I see. This situation can occur when HOCON parses value as a float? I think usually we will deal with the situations where some complex datatype (like IP or a comma-separated list) should be parsed from a string. In this case `hocon_schema` can use the default conversion, like you suggest here: https://github.com/emqx/hocon/pull/56#discussion_r618359516 , but that default conversion can benefit from inheriting the parser from the type.

Percent seems like a special case where a custom converter is required.

z8674558: ah I see. we can use custom converter in that situation.

z8674558: yes I agree that `(from|to)_string` is most commonly used.

k32: Sorry, took me a while to implement the prototype of the solution: https://github.com/k32/typerefl/pull/13

I forgot how painful Erlang metaprogramming is... :facepalm:

sample-schemas/demo_schema.erl
```diff
@@ -0,0 +1,58 @@
...
+
+setting(mapping) -> "app_foo.setting";
+setting(type) -> string();
+setting(converter) -> fun to_string/1;
+setting(override_env) -> "MY_OVERRIDE";
+setting(_) -> undefined.
+
+endpoint(mapping) -> "app_foo.endpoint";
+endpoint(type) -> ip();
+endpoint(converter) -> fun to_ip/1;
```
zmstone, line 27: `ip()` type should have a builtin converter so we do not have to write it here.
if this is only to demo how converter works, please add a comment.

sample-schemas/demo_schema.erl
zmstone: that's true.
but the namespace for source is almost always fixed?
if not, at leat we want it to be fixed?

I'd like to see a nested 'object' in the sample schema to trigger more discussion.
on top of my head:

```
fileds("foo") ->
  [{"f1", fun f1/1}];
fields("foo.nested") ->
  [{"nf1", fun nf1/1}].
```

zmstone: not all fields need a mapping target.
if the mapping clause returns `undefined`, it's treated as a value for the wrapping field.
or even a value for later retrieval when need for `translation`.

sample-schemas/demo_schema.erl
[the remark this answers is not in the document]
z8674558: applied

sample-schemas/demo_schema.erl
z8674558: we can have a built-in converter by https://github.com/k32/typerefl/pull/13 .

sample-schemas/demo_schema.erl
z8674558: > I'd like to see a nested 'object' in the sample schema to trigger more discussion.

added

sample-schemas/demo_schema.erl
k32: Tag `0.3.0` should support custom validators and parser. Example annotations: https://github.com/k32/typerefl/blob/master/test/transform_tests.erl#L22

src/hocon_schema.erl
```diff
@@ -0,0 +1,261 @@
...
+
+    % array
+    {ok, M6} = hocon:binary("foo.numbers=[1,2,3]", #{format => richmap}),
+    ?assertEqual([{["app_foo", "numbers"], [1, 2, 3]}],
+        map(demo_schema, M6)).
+
+env_test() ->
+    {ok, M0} = hocon:binary("foo.setting=hello", #{format => richmap}),
+    Envs0 = [{"EMQX_FOO__SETTING", "hi"}, {"HOCON_ENV_OVERRIDE_PREFIX", "EMQX_"}],
+    ?assertEqual([{["app_foo", "setting"], "hi"}], with_envs(fun map/2, [demo_schema, M0], Envs0)),
```
k32, line 220: This could be extracted to a helper function and reused

sample-schemas/demo_schema.erl
```diff
@@ -0,0 +1,67 @@
...
+
+-export([namespaces/0, fields/1]).
+
+namespaces() -> ["foo", "a.b"].
+
+fields("foo") ->
+    [ {"setting", fun setting/1}
+    , {"endpoint", fun endpoint/1}
+    , {"greet", fun greet/1}
+    , {"numbers", fun numbers/1}
```
zmstone, line 21: make sense to reference `"a.b"` here?
e.g. add another field`, {"ab", fun ab/1}`

```
ab(mapping) -> undefined; % so it's just a type checker for `foo.ab`
ab(type) -> fields("ab").
% if fields have logical connections, here is the place to validate
ab(validator) -> fun(V) -> maps:get(<<"f1">>, V) > maps:get(<<"f2">>, V) end;
ab(_) -> undefined.
```

zmstone: I'm not sure how `ab(type) -> typerefl:map()` should work.

zmstone: then there is also a need for

`ab(type) -> array(fields("ab"))`

sample-schemas/demo_schema.erl
zmstone: sorry, bad example. here again:

```
fileds("foo") ->
  [{"f1", fun f1/1},
   {"f2", fun nested/1}
   {"f3", fun array_of_nested/1}
  ];
fields("foo.nested") ->
  [{"nf1", fun nf1/1}].

nested(type) -> fileds("foo.nested");
...

array_field(type) -> array(fields("foo.nested"));
...
```

zmstone: again, I'm not sure how `typerefl` fits in.
should we use `typerefl` only for primitive types ?

sample-schemas/demo_schema.erl
z8674558: what do you mean by `ab(type) -> fields("ab").` ?

rebar.config
```diff
@@ -1,6 +1,8 @@
 {minimum_otp_vsn, "21.0"}.
 
-{deps, [{getopt, "1.0.1"}]}.
+{deps, [ {getopt, "1.0.1"}
+       , {typerefl, {git, "https://github.com/k32/typerefl", {tag, "0.3.0"}}}
```
k32, line 4: `0.3.1` supports IP address types from the box

sample-schemas/demo_schema.erl
zmstone: if we consider `fields(Name) -> [{"f1", fun f1/1}]` is to declare type of an 'object',
then `f1(type) -> fields(SubName)` means a nested 'object'

rebar.config
z8674558: applied

sample-schemas/demo_schema.erl
z8674558: `ref` type added.

sample-schemas/demo_schema.erl
```diff
@@ -0,0 +1,57 @@
...
+
+birthdays(mapping) -> "a.b.birthdays";
+birthdays(type) -> birthdays();
+birthdays(_) -> undefined.
+
+int(mapping) -> "a.b.some_int";
+int(type) -> integer();
+int(_) -> undefined.
+
+ref_x_y(mapping) -> undefined;
```
zmstone, line 55: I might have gave a bad example above. sorry.
we actually need a demo which shows how it works to map an object to a target,
so, please change it to some mapping target.
the `{"some_int", fun int/1}` field has no need for a mapping target, that's fine.

then in the test case, we'd expect something like:
```
% ?_assertEqual([], F("foo.ref_x_y={some_int = 1}")) % not this, but below:
?_assertEqual([{["foo", "ref_x_y"], #{"some_int" => 1}}], F("foo.ref_x_y={some_int = 1}"))
% or maybe 
?_assertEqual([{["foo", "ref_x_y"], [{"some_int", 1}]}], F("foo.ref_x_y={some_int = 1}"))
```

z8674558: the output of `map` API should be the list of mapped values. Since `foo.ref_x_y` is a raw key, `translate` API should be responsible for it when we want to get `"foo.ref_x_y={some_int = 1}"` ?

zmstone: yes, if `mapping` clause returns `undefined` then it should not be mapped.

what I am after is (if not to change `ref_x_y`, then maybe add another one):
`ref_j_k(mapping) -> "app_foo.jkopts";`
so the raw values parsed from hocon is mapped to `jkopts`.
i.e. a one to one mapping from `ref_j_k.*` to `app_foo.jkopts.*`.
with this, we do not have to write mapping for each field,
hence we can use built in types like `fun int/1` for the fields.

z8674558: I see. demo added

sample-schemas/demo_schema.erl
```diff
@@ -0,0 +1,66 @@
...
+
+-include_lib("typerefl/include/types.hrl").
+
+-type birthdays() :: [#{m := 1..12, d := 1..31}].
+
+-reflect_type([birthdays/0]).
+
+-export([namespaces/0, fields/1]).
+
+namespaces() -> ["foo", "a.b", "x.y"].
```
zmstone, line 11: we should perhaps only export root level, or re-usable namespaces here ?
if `x.y` and `j.k` are only internally used, there is no need to expose them.

src/hocon_schema_builtin.erl
```diff
@@ -0,0 +1,52 @@
...
+            Int
+    end;
+convert(Bin, Type) when is_binary(Bin) ->
+    from_string(binary_to_list(Bin), Type);
+convert(Other, _Type) ->
+    Other.
+
+from_string(Str, Type) when is_list(Str) ->
+    {?type_refl, Callbacks} = Type,
+    case maps:get(from_string, Callbacks, undefined) of
```
k32, line 42: Why not using `typerefl:from_string()`?

z8674558:
```
2) hocon_schema:env_test_/0:226
     Failure/Error: ?assertEqual([{[[97,112,112,95,102,111,111],[103,114,101,101,116]],[104,101,108,108,111]}], F ( "" , [ { "EMQX_FOO__GREET" , "hello" } ] ))
       expected: [{["app_foo","greet"],"hello"}]
            got: {error,
                     {errorlist,
                         [{error,
                              "Expected type: string(\"^hello$\")\nGot: hello\n"}]}}
```

It looks `from_string` with regexp types are not added, I hope this solves https://github.com/k32/typerefl/pull/16

k32: Good catch, it was a bug on my side.

z8674558: bumped to 0.4.0

sample-schemas/demo_schema.erl
```diff
@@ -0,0 +1,87 @@
...
+
+int(mapping) -> "a.b.some_int";
+int(type) -> integer();
+int(_) -> undefined.
+
+priv_int(type) -> integer();
+priv_int(_) -> undefined.
+
+ref_x_y(mapping) -> undefined;
+ref_x_y(type) -> {ref, fields("x.y")};
```
zmstone, line 82: maybe can be simplified as below?
```suggestion
ref_x_y(type) -> {ref, "x.y"};
```

zmstone:
looks good in general.
I'd like to once again bring up the binary vs. atom discussion though.

It's less of a concern if hocon lib parses keys into atoms when it is only to deal with config files.
however, the scope is now extended, we'll also allow REST APIs to POST config changes.

I'd propose to use `binary()` keys in HOCON parse results.
Only convert to `atom()` keys after schema check.

to make it easier when hand-crafting the schema,
we perhaps only need atoms for keys and namespaces.
the mapping target can be a list of atoms instead of dot separated strings.

e.g.
```
fields() -> [foo, a].

fields(foo) ->
    [ {setting, fun setting/1}
    ...;
fields(a) ->
    [{a, fun(type) -> {ref, b}; (_) -> undefined end}];
fields(b) ->
    [{birthdays, fun birthdays/1},
     {someint, fun some_int/1}
    ].

birthdays(mapping) -> [app_foo, a, b, birthdays];
birthdays(type) -> .....

some_int(mapping) -> [app_foo, a, b, someint];
some_int(type) -> ...
```

sample-schemas/demo_schema.erl
zmstone: the list / array support so far are only typerefl lists.
we may as well need a `{list, {ref, "x.y"}}` ?

sample-schemas/demo_schema.erl
```diff
@@ -2,7 +2,7 @@
 
 -include_lib("typerefl/include/types.hrl").
 
--type birthdays() :: [#{m := 1..12, d := 1..31}].
+-type birthdays() :: [map()].
```
z8674558, line 5: @zmstone `atom keys -> binary keys` applied. Now we cannot have detailed typing on `birthdays`, as `-type` does not allow binary map keys.

zmstone: ok.
we should probably not encourage objects without a schema anyway.

zmstone (approved):

PR #59, opened 2021-05-10
feat(hocon_cli): add escript generate command

z8674558:


src/hocon_cli.erl
```diff
@@ -18,39 +18,280 @@
...
+% @TODO add logger module for test
+-define(LOG, logger).
+-else.
+-define(LOG, logger).
+-endif.
+
+cli_options() ->
+%% Option Name, Short Code, Long Code, Argument Spec, Help Message
+    [
+    {help, $h, "help", undefined, "Print this usage page"}
```
k32, line 37: Nit: indentation

src/hocon_cli.erl
```diff
@@ -18,39 +18,280 @@
...
+
+    SuggestedLogLevel = list_to_atom(proplists:get_value(log_level, ParsedArgs)),
+    LogLevel = case lists:member(SuggestedLogLevel, [debug, info, notice, warning,
+        error, critical, alert, emergency]) of
+                   true -> SuggestedLogLevel;
+                   _ -> notice
+               end,
+    logger:remove_handler(default),
+    logger:add_handler(cuttlefish, logger_std_h,
+        #{config => #{type =>standard_error},
```
k32, line 78:
```suggestion
        #{config => #{type => standard_error},
```

src/hocon_cli.erl
```diff
@@ -18,39 +18,280 @@
...
+        io_lib:format("~s.~p.~s.~s.~s.~s.~s.~s",
+            [Filename,
+                Y,
+                zero_pad(M),
+                zero_pad(D),
+                zero_pad(HH),
+                zero_pad(MM),
+                zero_pad(SS),
+                Extension
+            ]).
```
k32, line 257: `io_lib:format` supports padding:

```
> io:format("~2..0b.~2..0b~n", [2, 1]).
02.01
```

src/hocon_cli.erl
```diff
@@ -18,39 +18,280 @@
...
+
+zero_pad(Integer) ->
```
k32, line 259: This function is not needed, `io_lib:format` suports padding.

src/hocon_cli.erl
```diff
@@ -18,39 +18,280 @@
...
+maybe_log_file_error(_, ok) ->
+    ok;
+maybe_log_file_error(Filename, {error, Reason}) ->
+    ?LOG:error("Error writing ~s: ~s", [Filename, file:format_error(Reason)]),
+    ok.
+
+filename_maker(Filename, Date, Extension) ->
+    {{Y, M, D}, {HH, MM, SS}} = Date,
+    _DestinationFilename =
+        io_lib:format("~s.~p.~s.~s.~s.~s.~s.~s",
```
k32, line 248: The result of `io_lib:format` typically needs to be wrapped in `lists:flatten`, because it returns an iolist

test/hocon_cli_tests.erl
```diff
@@ -0,0 +1,87 @@
...
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%--------------------------------------------------------------------
+
+-module(hocon_cli_tests).
+
+-include_lib("eunit/include/eunit.hrl").
+
+-define(assertPrinted(___Text),
+  begin
```
k32, line 22: Nit: I think this begin/end block doesn't have any effect here.

src/hocon_cli.erl
z8674558: very nice

src/hocon_cli.erl
```diff
@@ -18,39 +18,265 @@
...
+                     fun(_, _) -> ok end
+             end,
+
+    case hocon_schema:generate(Schema, Conf) of
+        %{error, _X} ->
+            % @TODO print error
+        %    error;
+        NewConfig ->
+            AppConfig = proplists:delete(vm_args, NewConfig),
+            VmArgs = stringify(proplists:get_value(vm_args, NewConfig)),
```
k32, line 188: Since the data is going to be written into a file, you can avoid `lists:flatten` and `string:join` calls here. `file:write_file` function accepts iolists as inputs, and it's a better data structure performance-wise, because you'll avoid extra list concatentations.

src/hocon_cli.erl
```diff
@@ -18,39 +18,265 @@
...
+                    {Destination, DestinationVMArgs};
+                {Err1, Err2} ->
+                    maybe_log_file_error(Destination, Err1),
+                    maybe_log_file_error(DestinationVMArgs, Err2),
+                    error
+            end
+
+    end.
+
+-spec prune(file:name_all(), integer()) -> ok.
```
k32, line 207:
```suggestion
-spec prune(file:name_all(), non_neg_integer()) -> ok.
```

src/hocon_cli.erl
```diff
@@ -18,39 +18,265 @@
...
+    Ext = filename:extension(Filename),
+    Base = hd(string:tokens(filename:basename(Filename, Ext), ".")),
+    Files =
+        lists:sort(filelib:wildcard(Base ++ ".*" ++ Ext, Path)),
+
+    delete([ filename:join([Path, F]) || F <- Files], MaxHistory),
+    ok.
+
+-spec delete(file:name_all(), integer()) -> ok.
+delete(Files, MaxHistory) when length(Files) =< MaxHistory ->
```
k32, line 222: This algorithm is `O(n^2)`. A better approach would be to calculate the length once, and then subtract from it on each iteration.

src/hocon_cli.erl
```diff
@@ -18,39 +18,265 @@
...
+    end,
+    delete(Files, MaxHistory).
+
+-spec maybe_log_file_error(
+    file:filename(), ok |
+    {error, file:posix()  %% copied from file:format_error/1
+    | badarg
+    | terminated
+    | system_limit
+    | { integer(), module(), term() }}) -> ok.
```
k32, line 238: This type signature looks cluttered. Move something to standalone `-type ` definitions?

src/hocon_cli.erl
```diff
@@ -18,39 +18,265 @@
...
+                              _ -> {[help], []}
+                          end,
+    {Command, ExtraArgs} = case {lists:member(help, ParsedArgs), Extra} of
+                               {false, []} -> {generate, []};
+                               {false, [Cmd | E]} -> {list_to_atom(Cmd), E};
+                               _ -> {help, []}
+                           end,
+    {Command, ParsedArgs, ExtraArgs}.
+
+%% @doc main method for generating erlang term config files
```
k32, line 66:
```suggestion
%% @doc Entry point of the script
```

src/hocon_cli.erl
z8674558: fixed (also added test for `prune`)

test/hocon_cli_tests.erl
```diff
@@ -0,0 +1,116 @@
...
+                 {actual, unicode:characters_to_list(___Output)}]})
+        end;
+        error ->
+            erlang:error({assertPrinted_failed,
+                [{module, ?MODULE},
+                 {line, ?LINE},
+                 {expected, ___Text},
+                 {reason, timed_out_on_receive}]})
+    end)())
+    end).
```
k32, line 42: Looks like an unmatched `end`?

src/hocon_cli.erl
k32: There is still a call to `length(Files)` on every iteration, `length` is O(n).

k32 (approved):
LGTM, apart from the `length` call in the guard of `hocon_cli:delete`

z8674558:
sry I missed your comment

src/hocon_cli.erl
z8674558: I thought it is ok to call `length` here as `length` is called only once. After comparing `length(Files)` and `MaxHistory`, `do_delete` deletes excess files and returns ok.

k32: Sorry, I missed the `do_delete` part.

k32 (approved):

PR #63, opened 2021-05-21
feat(hocon_schema): schema for single level fields

terry-xiaoyu:
Sometimes fields in hocon file many have only one level like:

```
{id: 123}
```

This PR adds the support for writing schema for the above example as:

```
structs() -> ["id"].
fields("id") ->
    [fun(mapping) -> "id";
        (type) -> integer();
        (_) -> undefined
     end].
```

### DIFF
--- a/sample-schemas/demo_schema.erl
+++ b/sample-schemas/demo_schema.erl
@@ -9,6 +9,8 @@
 
 -export([structs/0, fields/1, translations/0, translation/1]).
 
+-define(FIELD(NAME, TYPE), fun(mapping) -> NAME; (type) -> TYPE; (_) -> undefined end).
+
 structs() -> [foo, "a.b", person, "id"].
 translations() -> ["app_foo"].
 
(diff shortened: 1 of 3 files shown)


sample-schemas/demo_schema.erl
```diff
@@ -43,8 +43,16 @@ fields("x.y") ->
 
 fields("j.k") ->
     [ {"some_int", fun priv_int/1}
-    ].
+    ];
+
+fields(person) ->
+    [{id, fun(mapping) -> "person.id"; (type) -> "id"; (_) -> undefined end}];
 
+fields("id") ->
```
zmstone, line 51: how about
`-define(FIELD(NAME, TYPE), {NAME, fun(mapping) -> NAME; (type) -> TYPE; (_) -> undefined end}).`

z8674558: in emqx.schema I added helper function `ref`
https://github.com/emqx/hocon/pull/62/files#diff-c503cb08c8b7abf5ed5c983df29909f57f33a33ad139020f8b0022904fd6ef6fR1359

z8674558 (approved):

PR #75, opened 2021-06-10
feat: support atom key map

turtleDeng:


z8674558 (approved):

zmstone (changes requested):
add tests please

src/hocon_schema.erl
```diff
@@ -196,7 +196,12 @@ check_plain(Schema, Conf, Opts0) ->
 do_check(Schema, Conf, Opts) ->
     case map(Schema, Conf, structs(Schema), Opts) of
         {[], NewConf} ->
-            NewConf;
+            case maps:get(atom_key, Opts, false) of
```
zmstone, line 199: `atom_key` should be added to `opts()` type spec.

k32:
A comment explaining what atom_key_map function does would be nice. Right now it's a bit hard to say what this option does just by reading the code.

PR #83, opened 2021-06-14
Feat cli user decide file names

zmstone:
Prior to this PR, the hocon_cli's `generate` command prints `run_erl` boot command args to stdio and expect Erlang node's boot script to capture it and use it n `run_erl`.
The main reason I believe was that the `generate` command takes a system time and use it as a part of the generated file names. However the disadvantages of this approach are:
* Logs have to be printed to stderr
* Not shellcheck friendly as the captured stdout line has spaces, and must be split when used as `run erl`

Now the node boot script is forced to decide the next (new) name of the files to be generated by calling the `now_time` command.


z8674558 (approved):
nice

PR #95, opened 2021-07-05
feat: check_plain function support atom key

Rory-Z:


z8674558 (dismissed):

src/hocon_schema.erl
```diff
@@ -365,6 +365,8 @@ str(S) when is_list(S) -> S.
 bin(A) when is_atom(A) -> atom_to_binary(A, utf8);
 bin(S) -> iolist_to_binary(S).
 
+atom(B) when is_binary(B) -> binary_to_atom(B, utf8).
```
zmstone, line 368: to existing atom ?

zmstone (changes requested):

src/hocon_schema.erl
Rory-Z: Fixed

src/hocon_schema.erl
```diff
@@ -767,7 +770,10 @@ do_plain_get([], Conf) ->
     %% value as-is
     Conf;
 do_plain_get([H | T], Conf) ->
-    Child = maps:get(H, Conf, undefined),
+    Child = case maps:get(H, Conf, undefined) of
```
zmstone, line 773: Catch error if it’s not a an existing atom, return undefined

zmstone:
Please ensure 100% test coverage

src/hocon_schema.erl
```diff
@@ -767,7 +770,10 @@ do_plain_get([], Conf) ->
     %% value as-is
     Conf;
 do_plain_get([H | T], Conf) ->
-    Child = maps:get(H, Conf, undefined),
```
zmstone, line 770: Not this pr’s issue, but: we need another clause to handle when Conf is not a map.

PR #97, opened 2021-07-08
fix(hocon_schema): add test for validating array of primitives

terry-xiaoyu:
`hocon_schema:check/2` failed on an array of string(), if the default value is applied.


zmstone:
fixed here: https://github.com/emqx/hocon/pull/101/files

PR #108, opened 2021-07-14
feat(hocon_schema): support set array value using env

terry-xiaoyu:
Just want to set the array values using environment, like this:

```
EMQX_CLUSTER__STATIC__SEEDS="[emqx@node1.emqx.io, emqx@node2.emqx.io]" sh -x _build/emqx/rel/emqx/bin/emqx console
```

### DIFF
--- a/src/hocon_schema.erl
+++ b/src/hocon_schema.erl
@@ -568,14 +568,14 @@ resolve_field_value(Schema, FieldValue, Opts) ->
         undefined ->
             resolve_default_override(Schema, FieldValue, Opts);
         EnvValue ->
-            boxit(Opts, EnvValue, FieldValue)
+            maybe_mkrich(Opts, EnvValue, ?EMPTY_BOX)
     end.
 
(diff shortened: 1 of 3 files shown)


src/hocon_schema.erl
```diff
@@ -599,15 +599,29 @@ collect_envs() ->
...
+read_hocon_val(Value) ->
+    try {ok, HoconVal} = hocon:binary(Value, #{}),
+        HoconVal
+    catch
+        _ : _ -> read_informal_hocon_val(Value)
+    end.
+
+read_informal_hocon_val(Value) ->
+    BoxedVal = "virtual_root=" ++ Value,
+    {ok, HoconVal} = hocon:binary(BoxedVal, #{}),
```
zmstone, line 615: raise exception when failed to parse this value

terry-xiaoyu: fixed

zmstone (approved):

PR #113, opened 2021-07-20
fix(hocon_schema): remove the warning log when parse env value failed

terry-xiaoyu:


src/hocon_schema.erl
```diff
@@ -614,12 +614,7 @@ read_informal_hocon_val(Value, Opts) ->
     case hocon:binary(BoxedVal, #{}) of
         {ok, HoconVal} ->
             maps:get(<<"fake_key">>, HoconVal);
-        {error, Reason} ->
-            Msg = iolist_to_binary(
-                    io_lib:format(
-                      "invalid_hocon_string: ~p,reason: ~p",
-                      [Value, Reason])),
-            log(Opts, warning, Msg),
+        {error, _Reason} ->
```
zmstone, line 617: maybe change to debug

terry-xiaoyu: fixed

zmstone (approved):

PR #137, opened 2021-09-14
feat(options): support env_override options to skip env override

zhongwencool:
For those who (frequently) call the hocon_schema:check function, provide an option to have it not check the env from os.

I check schema frequently by minirest api but without env.


zmstone (approved):

zmstone:
tagged `0.18.0`

PR #138, opened 2021-09-16
feat(hoconsc): is_type/1 to check if input is a hocon type

zhongwencool:
Check if args is a hocon type.
Use `hocon_schema:field_schema(#{"bar" => foo}, good) ` will crash.
so I need check the input is a hocon type first.

### DIFF
--- a/src/hoconsc.erl
+++ b/src/hoconsc.erl
@@ -21,6 +21,7 @@
 -export([ref/1, ref/2]).
 -export([array/1, union/1, enum/1]).
 -export([lazy/1, map/2]).
+-export([is_type/1]).
 
 -include("hoconsc.hrl").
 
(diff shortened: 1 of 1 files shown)


k32 (dismissed):

src/hoconsc.erl
```diff
@@ -48,10 +49,22 @@ union(OfTypes) when is_list(OfTypes) -> ?UNION(OfTypes).
...
+%% @doc make a map type.
 map(Name, Type) -> ?MAP(Name, Type).
 
+%% @doc Check Type is a hocon type.
+is_type(?UNION(Members)) -> lists:all(fun is_type/1, Members);
+is_type(?ARRAY(ElemT)) -> is_type(ElemT);
+is_type(?LAZY(HintT)) -> is_type(HintT);
+is_type(?REF(_)) -> true;
+is_type(?R_REF(_, _)) -> true;
+is_type(#{type := _}) -> true;
```
zmstone, line 64: schema includes type.
schema can be a map or an anonymous function.
From a schema one can retrieve either type or other metadata

maybe name it `is_schema` instead

zhongwencool: done @zmstone

zmstone (approved):

PR #144, opened 2021-09-22
fix(schema): do not validate if no_conversion for non-primtive

zmstone:
in case of `only_fill_defaults`, we do not keep the converted value
in returned results when validating recursively.

for non-primtive values, there is no way to validate them, because the validator callbacks expect converted value.

### DIFF
--- a/src/hocon_schema.erl
+++ b/src/hocon_schema.erl
@@ -99,7 +99,11 @@
 -define(FROM_ENV_VAR(Name, Value), {'$FROM_ENV_VAR', Name, Value}).
 -define(IS_NON_EMPTY_STRING(X), (is_list(X) andalso X =/= [] andalso is_integer(hd(X)))).
 -type loggerfunc() :: fun((atom(), map()) -> ok).
+%% Config map/check options.
 -type opts() :: #{ logger => loggerfunc()
+                   %% no_conversion is to only to fill default values for the input
+                   %% config to be checked, only primitive value type check (validation)
+                   %% but not complex value validation and mapping
                  , no_conversion => boolean()
                  , atom_key => boolean()
                  , return_plain => boolean()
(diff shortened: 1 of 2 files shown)


src/hocon_schema.erl
```diff
@@ -600,14 +618,22 @@ map_field(Type, Schema, Value0, Opts) ->
...
         _ -> {ValidationResult, boxit(Opts, ConvertedValue, Value0)}
     end.
 
+is_primitive_type(?ARRAY(_)) -> false;
+is_primitive_type(?LAZY(_)) -> false;
+is_primitive_type(?UNION(_)) -> false;
+is_primitive_type(?MAP(_, _)) -> false;
+is_primitive_type(?REF(_)) -> false;
+is_primitive_type(?R_REF(_, _)) -> false;
+is_primitive_type(_) -> true.
```
qzhuyan, line 635: Why not match primitive_types instead? defaults to true feels wrong?

src/hocon_schema.erl
```diff
@@ -505,10 +509,24 @@ map_fields([{FieldName, FieldSchema} | Fields], Conf0, Acc, Opts) ->
...
+            true ->
+                %% primitive values are already validated
+                [];
+            false ->
+                %% otherwise valdiate using the schema defined callbacks
+                validators(field_schema(FieldSchema, validator))
+        end,
     case find_errors(Acc) of
+        ok when NoConversion ->
+            %% when no_conversion, we are only filling default values (recursively)
```
qzhuyan, line 524: I don't have much context here, 
`no_conversion` means use default values? Shall we change it to `use_default_val`?

zmstone: changed the option to `only_fill_defaults`

qzhuyan (approved):

PR #145, opened 2021-10-07
feat(schema): Add structs doc

zmstone:
Support document generation for struct (type)

### DIFF
--- a/src/hocon_schema.erl
+++ b/src/hocon_schema.erl
@@ -21,10 +21,12 @@
 %% behaviour APIs
 -export([ roots/1
         , fields/2
+        , fields_and_meta/2
         , translations/1
         , translation/2
         , validations/1
         ]).
+
 
 -export([map/2, map/3, map/4]).
 -export([translate/3]).
@@ -33,7 +35,8 @@
 -export([richmap_to_map/1, get_value/2, get_value/3]).
 -export([namespace/1, resolve_struct_name/2, root_names/1]).
 -export([field_schema/2, override/2]).
--export([find_structs/1]).
+
+-export([find_structs/1]). %% internal use
 
 -include("hoconsc.hrl").
 -include("hocon_private.hrl").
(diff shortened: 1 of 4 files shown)


k32 (approved):

src/hocon_schema.erl
```diff
@@ -84,13 +87,17 @@
          , desc => iodata()
          }.
 
+-type struct_meta() :: #{desc => iodata()}.
 -type field() :: {name(), typefunc() | field_schema()}.
+-type fields() :: [field()] | #{fields := [field()],
```
zhongwencool, line 92: There are more meta's to be added than just `desc`, such as: 
I need to set this meta to mark this configuration as hot updatable, generating the HTTP API put method.

src/hocon_schema.erl
```diff
@@ -194,13 +201,23 @@ find_structs(Schema) ->
...
 
+%% @doc Get fields of the struct for the given struct name.
 -spec fields(schema(), name()) -> [field()].
-fields(Mod, Name) when is_atom(Mod) ->
-    Mod:fields(Name);
-fields(#{fields := Fields}, Name) when is_function(Fields) ->
-    Fields(Name);
-fields(#{fields := Fields}, Name) when is_map(Fields) ->
-    maps:get(Name, Fields).
+fields(Sc, Name) ->
```
zhongwencool, line 206: if we only return fields meta by `fields/1`, then the `find_struct/1` can't be private. 
I need the get struct from emqx_dashboard apps to generate API.

so `find_struct/1` should be public or make fields/1 return all meta(bad idea...)

zmstone: There is fields_and_meta/1

zmstone: Exported

zhongwencool (dismissed):

zmstone:
rebased

PR #160, opened 2021-11-29
refactor: apply env for array elements

zmstone:
before this change, `array` elements can not be overridden by environment variables
the work around was to use `map("id", ElementType)` type instead.
for instance, if we want to override the first element of an array `["element1", "element2"]` to `"elementA"`,
we'd have work with `#{<<"1">> => "element1", <<"2">> => "element2"}`
so that the elements can be set as `export ENV_NAMESPACE__1="elementA"`.

this PR made it possible to override array elements directly.

### DIFF
--- a/src/hocon_schema.erl
+++ b/src/hocon_schema.erl
@@ -432,6 +432,11 @@ map(Schema, Conf0, Roots0, Opts0) ->
     ok = assert_integrity(Schema, NewConf, Opts),
     {Mapped, maybe_convert_to_plain_map(NewConf, Opts)}.
 
+%% Merge environment overrides into HOCON value before checking it against the schema.
+%% An alternative implimentation is to read environment variables while checking the
+%% fields, however it is more complicated than the current approach
+%% because, for nullable fields, we may skip over the struct schema,
+%% meaning override only works if the object existed in the input Conf.
 apply_envs(_EnvNamespace, _Envs, _Opts, [], Conf) -> Conf;
 apply_envs(EnvNamespace, Envs, Opts, [{RootName, RootSc} | Roots], Conf) ->
     ShouldApply =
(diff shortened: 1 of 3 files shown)


src/hocon_schema.erl
```diff
@@ -432,6 +432,11 @@ map(Schema, Conf0, Roots0, Opts0) ->
     ok = assert_integrity(Schema, NewConf, Opts),
     {Mapped, maybe_convert_to_plain_map(NewConf, Opts)}.
 
+%% Merge environment overrides into HOCON value before checking it against the schema.
+%% An alternative implimentation is to read environment variables while checking the
+%% fields, however it is more complicated than the current approach
+%% because for nullable fields, we may skip over the struct schema,
```
k32, line 438: Nit:  because for -> because of

src/hocon_schema.erl
```diff
@@ -1059,13 +1091,64 @@ deep_put(Opts, Path, Value, Conf) ->
...
+
+check_index_seq(_, [], Acc) ->
+    {ok, lists:reverse(Acc)};
+check_index_seq(I, [{I, V} | Rest], Acc) ->
+    check_index_seq(I + 1, Rest, [V | Acc]);
+check_index_seq(I, [{Unexpected, _} | _Rest], _) ->
+    {error, #{expected_index => I,
+              got_index => Unexpected}}.
+
+type_hint(B) when is_binary(B) -> string; %% mayebe secret, do not hint value
```
k32, line 1149: Nit (typo): maybe

k32 (dismissed):

src/hocon_schema.erl
zmstone: maybe "because, for nullable fields" ?

zmstone:
fixed elvis check and typo in the latest force push

src/hocon_schema.erl
k32: Aha, nevermind my comment then.

k32 (approved):

PR #172, opened 2022-01-23
refactor: moved deep_merge/2 to hocon_maps and deleted deep_map_merge/2

zmstone:
deep_merge/2 is generic enough for deep_map_merge/2
deep_map_merge was to merge map array elements when the array
from the second arg is a indexed-array

all the existing cases where deep_map_merge was used had no
indexed-array

### DIFF
--- a/src/hocon_maps.erl
+++ b/src/hocon_maps.erl
@@ -17,7 +17,8 @@
 -module(hocon_maps).
 
 %% Deep ops of deep maps.
--export([deep_get/2, deep_put/4]).
+-export([deep_get/2, deep_put/4,
+         deep_merge/2]).
 
 %% Access maybe-rich map values,
 %% Always reutrn plain value.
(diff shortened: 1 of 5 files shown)


src/hocon_maps.erl
```diff
@@ -146,3 +147,92 @@ try_get(Key, Conf, map) when is_list(Conf) ->
...
+do_update_array_element(List, Index, GoDeep) when is_list(List) ->
+    MinIndex = 1,
+    MaxIndex = length(List) + 1,
+    Index < MinIndex andalso throw({bad_array_index, "index starts from 1"}),
+    Index > MaxIndex andalso
+    begin
+        Msg0 = io_lib:format("should not be greater than ~p.", [MaxIndex]),
+        Msg1 = case Index > 9 of
+                   true ->
+                       "~nEnvironment variable overrides applied in alphabetical "
```
thalesmg, line 222: Small nit: _alphabetical order_

thalesmg (approved):

src/hocon_maps.erl
zmstone: thank you. will fix in the next PR.

PR #182, opened 2022-03-18
Fix: schema spec

zhongwencool:
1. schema can define as :
```erlang
#{roots => [{"val", hoconsc:map(key, string())}]},
```
the `fields` is not required.

2. add ?R_REF/1 ?MK/1 ?MK/2 macro:
we do a lot of import actions like:
```erlang
-import(hoconsc, [mk/1, mk/2, ref/1, ref/2, array/1, enum/1]).
```
we can use macro to replace the `import`.

### DIFF
--- a/include/hoconsc.hrl
+++ b/include/hoconsc.hrl
@@ -14,16 +14,17 @@
 %% limitations under the License.
 %%--------------------------------------------------------------------
 
-%% This header file is inteded for internal use in schema modules.
-
 -ifndef(HOCONSC_HRL).
 -define(HOCONSC_HRL, true).
+
+-include_lib("typerefl/include/types.hrl").
 
 -define(ARRAY(OfTYpe), {array, OfTYpe}).
 -define(UNION(OfTypes), {union, OfTypes}).
 -define(ENUM(OfSymbols), {enum, OfSymbols}).
 -define(REF(Name), {ref, Name}).
 -define(R_REF(Module, Name), {ref, Module, Name}). % remote ref
+-define(R_REF(NAME), ?R_REF(?MODULE, NAME)).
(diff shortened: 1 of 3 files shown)


include/hoconsc.hrl
```diff
@@ -34,4 +35,8 @@
 %% Map keys are always strings, `Name' is only for documentation
 -define(MAP(Name, Type), {map, Name, Type}).
 
+%% To avoid not import those function. we provide a macro to call them.
+-define(MK(Type), hoconsc:mk(Type)).
+-define(MK(Type, Meta), hoconsc:mk(Type, Meta)).
```
terry-xiaoyu, line 40: Should we reload all of modules relies on this header file, if it changed and release upgrade is needed?

include/hoconsc.hrl
```diff
@@ -34,4 +35,8 @@
 %% Map keys are always strings, `Name' is only for documentation
 -define(MAP(Name, Type), {map, Name, Type}).
 
+%% To avoid not import those function. we provide a macro to call them.
+-define(MK(Type), hoconsc:mk(Type)).
```
zmstone, line 39: `MK` is probably not a good name.
Maybe simply `HOCON`

include/hoconsc.hrl
zhongwencool: I think we don't need to change this macro anymore, we only change the function implementation `hoconsc:mk/2`. This macro is just a shortcut to a function, too simple to be modified in the future.
The only case I can think of where it might be modified is when `hoconsc:mk/2` function is renamed. But we should add new function to do that instead of changing the name.

zmstone (approved):

PR #184, opened 2022-03-24
fix(schema): no need to check format when converting to plain map

zmstone:
The rich-map to plain-map conversion used to rely on the `format`
option. Later is_richmap was introduced, so there is no need to
check option any more.

This commit also fixes a bug in is_richmap:
the boxing layer might be unboxed, so it need to go at least
one level down to check if there is a `$hcVal` field.


thalesmg (approved):

zmstone:
0.26.4

PR #185, opened 2022-03-25
feat: Add desc callback.

ieQu1:


zmstone (approved):

src/hocon_schema.erl
```diff
@@ -197,12 +200,30 @@ fields(Sc, Name) ->
 %% @doc Get fields and meta data of the struct for the given struct name.
 -spec fields_and_meta(schema(), name()) -> fields().
 fields_and_meta(Mod, Name) when is_atom(Mod) ->
-    ensure_struct_meta(Mod:fields(Name));
+    case Mod:fields(Name) of
+        Fields when is_list(Fields) ->
+            maybe_add_desc(Mod, Name, #{fields => Fields});
+        Fields ->
+            ensure_struct_meta(Fields)
```
thalesmg, line 207: Q: shouldn't the description be added with `maybe_add_desc/3` in this case as well?

thalesmg (approved):
Suggestion: maybe create a test that asserts that `desc` ends up in the expected place(s)?

src/hocon_schema.erl
ieQu1: No, this case handles a different (older) style of adding descriptions.

PR #187, opened 2022-04-12
feat: desc support struct.

zhongwencool:


src/hocon_schema.erl
```diff
@@ -350,3 +351,15 @@ assert_unique_field_names([{Name, _} | Rest], Acc) ->
         false ->
             assert_unique_field_names(Rest, [BinName | Acc])
     end.
+
+unbox_desc(?DESC(Mod, Id)) ->
```
zmstone, line 355:
```suggestion
resolve_desc(?DESC(Mod, Id)) ->
```

src/hocon_schema.erl
```diff
@@ -324,9 +325,9 @@ field_schema(?UNION(_) = Union, SchemaKey) ->
 field_schema(?ENUM(_) = Enum, SchemaKey) ->
     field_schema(hoconsc:mk(Enum), SchemaKey);
 field_schema(FieldSchema, SchemaKey) when is_function(FieldSchema, 1) ->
-    FieldSchema(SchemaKey);
+    unbox_desc(FieldSchema(SchemaKey));
 field_schema(FieldSchema, SchemaKey) when is_map(FieldSchema) ->
-    maps:get(SchemaKey, FieldSchema, undefined).
+    unbox_desc(maps:get(SchemaKey, FieldSchema, undefined)).
```
zmstone, line 330: putting desc resolution here feels too low level and unnecessary expensive as it has to load hocon for each and every filed.

it should be enough to resolve it in hocon_schema_json and hocon_schema_md.

zhongwencool: All needs desc should be resolved.
emqx swagger's desc also need.
if we expose the resolve function to the user. user must call:
```erlang
Desc = hocon_schema:field_schema(Hocon,desc),
hocon_schema:resolve_desc(Desc).
```
it only resolves the desc field, it is not expensive for other fields(we only match `?DESC`)

zhongwencool: but the other fields call `resolve_desc/1` if not necessary.
I want to introduce a new function called hocon_schema:desc/1.
```erlang
hocon_schema:desc(Hocon) ->
Desc = hocon_schema:field_schema(Hocon,desc),
hocon_schema:resolve_desc(Desc).
```
If you want to get the desc, no longer use `field_schema/2` directly.

zmstone: OK.
But maybe `resolve_schema(ResolverFun)`, and `ResolverFun` can have a cache implemented inside.

terry-xiaoyu (approved):

src/hocon_schema.erl
```diff
@@ -350,3 +357,36 @@ assert_unique_field_names([{Name, _} | Rest], Acc) ->
...
+      ets:insert(Tab, {File, Schema}),
+      #{tab => Tab, file => File};
+    {error, Reason} ->
+      error(#{reason => Reason, file => File})
+  end.
+
+delete_cache(#{tab := undefined}) -> ok;
+delete_cache(#{tab := Tab}) -> ets:delete(Tab).
+
+resolve_schema(?DESC(Mod, Id), State) ->
```
zmstone, line 375: resolve_desc is a better name?

src/hocon_schema.erl
```diff
@@ -350,3 +357,36 @@ assert_unique_field_names([{Name, _} | Rest], Acc) ->
         false ->
             assert_unique_field_names(Rest, [BinName | Acc])
     end.
+
+new_cache(undefined) -> #{file => undefined, tab => undefined};
```
zmstone, line 361: New_desc_cache?

src/hocon_schema_html.erl
```diff
@@ -26,57 +26,60 @@
 -define(REF_PREFIX_ROOT, "root-").
 -define(REF_PREFIX_STRUCT, "struct-").
 
-gen(Schema, Title) ->
+gen(Schema, Title, File) ->
     {RootNs, RootFields, Structs} = hocon_schema:find_structs(Schema),
     IndexHtml = fmt_index(RootFields, Structs),
+    Cache = hocon_schema:new_cache(File),
```
zmstone, line 32: `tru`
And delete cache in `after` clause?

src/hocon_schema_json.erl
```diff
@@ -23,20 +23,31 @@
...
+    {File, Opts1} = maps:take(desc_file, Opts),
+    Cache = hocon_schema:new_cache(File),
+    Opts2 = Opts1#{cache => Cache},
+    Json =
+      [ gen_struct(RootNs, RootNs, "Root Config Keys", #{fields => RootFields}, Opts2)
+        | lists:map(fun({Ns, Name, Fields}) ->
+        gen_struct(RootNs, Ns, Name, Fields, Opts2)
+                    end, Structs)
+      ],
+    hocon_schema:delete_cache(Cache),
```
zmstone, line 49: Do it in `try after`

src/hocon_schema_html.erl
```diff
@@ -26,57 +26,60 @@
 -define(REF_PREFIX_ROOT, "root-").
 -define(REF_PREFIX_STRUCT, "struct-").
 
-gen(Schema, Title) ->
+gen(Schema, Title, File) ->
```
zmstone, line 29: If it’s a per schema module desc file. Passing in from here won’t work?
The cache insert will have to be dynamic while traversing the schema tree.

src/hocon_schema.erl
zhongwencool: we already have `field_schema(Schema,Key)`, 
so we can do `field_schema(Schema,default), field_schema(Schema, required) etc`
we don't choose `get_required_field(Schema)`, this make too many functions.

so I don't think resolve_desc is a better name, It may be resolve_xxx in the future.
Maybe `resolve_field`?

src/hocon_schema_html.erl
zhongwencool: Current only one desc file support.
if we support  "per schema module desc file".
We have to declare the name of each desc implicitly.
we need to declare as :
```erlang
?DESC("path/to/desc/file", Id).
```
otherwise, we need to put all desc files in the same dir. and define
```erlang
?DESC("DescFile", Id).
```

Neither of the above 2 solutions is as easy as merging the desc directly into a large file.

zmstone: understood, please document it for the api?

src/hocon_schema.erl
zmstone: Got it. resolve_schema makes sense.

PR #195, opened 2022-04-24
fix: struct order error if namespace is undefined

zhongwencool:
If the namespace of multiple modules is undefined(the same namespace), the order of structs will be mixed up.

### DIFF
--- a/src/hocon_schema.erl
+++ b/src/hocon_schema.erl
@@ -300,8 +300,8 @@ find_structs_per_type(_Schema, _Type, Acc, _Stack, _TStack) ->
     Acc.
 
 find_ref(Schema, Name, Acc, Stack, TStack) ->
-    Namespace = hocon_schema:namespace(Schema),
-    Key = {Namespace, Name},
+    Namespace = namespace(Schema),
+    Key = {Namespace, Schema, Name},
     Path = path(Stack),
     Paths =
         case maps:find(Key, Acc) of
@@ -327,7 +327,7 @@ find_structs(Schema) ->
     RootNs = hocon_schema:namespace(Schema),
     {RootNs, RootFields, [
         {Ns, Name, Fields}
-     || {{Ns, Name}, Fields} <- lists:keysort(1, maps:to_list(All))
+     || {{Ns, _Schema, Name}, Fields} <- lists:keysort(1, maps:to_list(All))
     ]}.
 
 unify_roots(Schema) ->
--- a/src/hocon_tconf.erl
+++ b/src/hocon_tconf.erl
@@ -1134,7 +1134,7 @@ check_index_seq(I, [{Index, V} | Rest], Acc) ->
 
 get_invalid_name(Names) ->
     lists:filter(
-        fun(F) ->            
+        fun(F) ->
             nomatch =:=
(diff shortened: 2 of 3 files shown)


src/hocon_schema.erl
```diff
@@ -300,8 +300,8 @@ find_structs_per_type(_Schema, _Type, Acc, _Stack, _TStack) ->
     Acc.
 
 find_ref(Schema, Name, Acc, Stack, TStack) ->
-    Namespace = hocon_schema:namespace(Schema),
-    Key = {Namespace, Name},
+    Namespace = namespace(Schema),
+    Key = {Schema, Namespace, Name},
```
zmstone, line 304: The lookup key is ok to be only namespace because Schema is not necessary always an atom

src/hocon_schema.erl
```diff
@@ -327,7 +327,7 @@ find_structs(Schema) ->
     RootNs = hocon_schema:namespace(Schema),
     {RootNs, RootFields, [
         {Ns, Name, Fields}
-     || {{Ns, Name}, Fields} <- lists:keysort(1, maps:to_list(All))
+     || {{_Schema, Ns, Name}, Fields} <- lists:keysort(1, maps:to_list(All))
```
zmstone, line 330: Fix the problem with a customized sort function instead.

zmstone (changes requested):

src/hocon_schema.erl
zmstone: Or: define namespace for all!

zhongwencool: I think defining a namespace for every schema module is too strict.
we should allow an undefined namespace.

src/hocon_schema.erl
zhongwencool: schema is able to sort even not an atom.

src/hocon_schema.erl
zhongwencool: I already defined the namespace for all here https://github.com/emqx/emqx/pull/7742
But we should also allow if the namespace is the same, the sort by schema.

src/hocon_schema.erl
```diff
@@ -327,7 +327,7 @@ find_structs(Schema) ->
     RootNs = hocon_schema:namespace(Schema),
     {RootNs, RootFields, [
         {Ns, Name, Fields}
-     || {{Ns, Name}, Fields} <- lists:keysort(1, maps:to_list(All))
+     || {{Ns, _Schema, Name}, Fields} <- lists:keysort(1, maps:to_list(All))
```
zmstone, line 330: If we call this, there is no need to have schema in the key
```
sort(Fun, List1) -> List2
Types

Fun = fun((A :: T, B :: T) -> [boolean()](https://www.erlang.org/doc/man/erlang.html#type-boolean))
List1 = List2 = [T]
T = [term()](https://www.erlang.org/doc/man/erlang.html#type-term)
```

zmstone: Or sort by display name . In binary format

zmstone (approved):

zmstone (approved):

PR #208, opened 2022-07-08
EMQX-6018 add check_schema command

mononym:


src/hocon_cli.erl
```diff
@@ -341,6 +344,26 @@ generate(ParsedArgs) ->
             stop_deactivate()
     end.
 
+-spec check_schema([proplists:property()]) -> ok.
+check_schema(ParsedArgs) ->
+    Schema = load_schema(ParsedArgs),
+    Conf = load_conf(ParsedArgs, fun log/3),
+    LogFun =
+        case proplists:get_value(verbose_env, ParsedArgs) of
```
zmstone, line 352: this is perhaps never be set to `true` for check_schema.

zmstone (commented):
looks good. only one minor

zmstone (approved):

PR #209, opened 2022-08-09
feat: add field deprecation

zmstone:
related: https://github.com/emqx/emqx/issues/8672

The field schema is kept, so the old configuration will continue to be checked OK.
Deprecated field value will be dropped from type-check result,
so the application code should not need to deal with unexpected fields.

### DIFF
--- a/src/hocon_schema.erl
+++ b/src/hocon_schema.erl
@@ -38,7 +38,8 @@
     path/1,
     readable_type/1,
     fmt_type/2,
-    fmt_ref/2
+    fmt_ref/2,
+    is_deprecated/1
 ]).
 
 -export([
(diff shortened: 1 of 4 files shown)


src/hocon_schema.erl
```diff
@@ -119,6 +119,12 @@
         desc => desc(),
         %% hide it from doc generation
         hidden => boolean(),
+        %% this field is deprecated,
+        %% but can not be removed due to compatibility reasons
+        %% this filed will not appear in generated docs.
```
thalesmg, line 124:
```suggestion
        %% this field will not appear in generated docs.
```

test/hocon_tconf_tests.erl
```diff
@@ -720,9 +720,33 @@ required_test() ->
...
+            Sc,
+            #{<<"f2">> => <<"string">>},
+            #{required => false}
+        )
+    ),
+    ?assertEqual(
+        #{<<"f2">> => "string"},
+        hocon_tconf:check_plain(
+            Sc,
+            #{<<"f2">> => <<"string">>, <<"f3">> => 4444},
```
thalesmg, line 745: Suggestion: maybe test that we can even change the type of the values and that won't be validated?

thalesmg (dismissed):

thalesmg (approved):

PR #212, opened 2022-11-30
fix(tconf): allow converter to return 'undefined'

zmstone:
Converter returning 'undefined' has no effect, but should not crash.


thalesmg (approved):

zmstone:
rebasaed

PR #214, opened 2022-12-06
feat(tags): add support for tagging schemas

thalesmg:
[EMQX-8330](https://emqx.atlassian.net/browse/EMQX-8330)

By adding tags, it'll be (hopefully) easier to manipulate and split a big schema that has been dumped to JSON based on those.

### DIFF
--- a/src/hocon_schema_md.erl
+++ b/src/hocon_schema_md.erl
@@ -16,7 +16,7 @@
 
 -module(hocon_schema_md).
 
--export([gen/2]).
+-export([gen/2, gen_from_structs/2]).
 
 -include("hoconsc.hrl").
 -include("hocon_private.hrl").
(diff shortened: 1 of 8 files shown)


src/hocon_schema_md.erl
```diff
@@ -129,6 +132,8 @@ fmt_mapping(Str) -> ["`", Str, "`"].
 
 fmt_type(T) -> hocon_md:code(do_type(T)).
 
+do_type(#{kind := <<Kind/binary>>} = Type) ->
```
zmstone, line 135: not quite following why is this needed.

thalesmg: When one reads the schema from an existing JSON file, the fields and values are binaries, so this was blowing up.

It didn't blow up before the refactoring because the schema was processed in-memory, without being written first to disk.  So the key/values stayed as atoms.

(at least that's what I remember, it has been some weeks).

zmstone (approved):

PR #226, opened 2023-01-17
fix: badmap crash when error context is not a map

zmstone:


src/hocon_tconf.erl
```diff
@@ -1295,13 +1295,10 @@ maybe_hd(Other) -> Other.
 readable_type(T) ->
     str(hocon_schema:readable_type(T)).
 
+ensure_type_path(#{matched_type := SubType} = ErrorContext, ReadableType) ->
+    ErrorContext#{matched_type => ReadableType ++ "/" ++ SubType};
```
thalesmg, line 1299: I guess it's not possible for `SubType` to be `""`?

thalesmg (dismissed):

src/hocon_tconf.erl
zmstone: no.

thalesmg (approved):

PR #233, opened 2023-03-01
feat: skip calling validators on `hoconsc:map`'s

thalesmg:
Currently, if we define a field which is a `hoconsc:map` and define a validator for it, the validator function is called once for each of the map's fields, and also on the map as a whole after each field is checked.

Since the keys of the map are always just a (string/atom) name and each of the fields already have the validators called on them, we should skip calling such validators from maps on the whole map itself.

Calling the validators on the whole map is particularly inconvenient specially in the case in which the fields are themselves structs.  In that case, it's hard for the validator function to differentiate between been called with a field or the whole map.


zmstone:
what about the non-struct values ? e.g. `map(integer())`

PR #234, opened 2023-03-13
feat: change env's invalid hocon value from string to binary

zhongwencool:
keep ENV's invalid hocon value as binary when we store raw config.
before:
```bash
# before
EMQX_STATSD__SERVER='127.0.0.1:8124' ./bin/emqx console
> emqx:get_raw_config([statsd,server]).
"127.0.0.1:8124"
# after
> emqx:get_raw_config([statsd,server]).
<<"127.0.0.1:8124">>
```

Report from : [StatsD server address changed to array format](https://emqx.atlassian.net/browse/EMQX-9154)


qzhuyan (approved):

qzhuyan:
LGTM, I think it matched what we decided 2 weeks ago.

thalesmg (approved):

PR #235, opened 2023-03-13
feat: add struct (root included) field name aliases

zmstone:
With aliases supported, we can more freely rename fileds. e.g.
if a field is renamed to `new_field`, all we need to do is to keep the old name `field` as an alias in the schema like this:
`{new_filed, #{type => integer(), aliases => [field]}`

### DIFF
--- a/src/hocon_tconf.erl
+++ b/src/hocon_tconf.erl
@@ -14,6 +14,7 @@
 %% limitations under the License.
 %%--------------------------------------------------------------------
 
+%% tconf: typed-config
 -module(hocon_tconf).
 
 -elvis([{elvis_style, god_modules, disable}]).
@@ -458,6 +459,8 @@ map_fields_cont([{_, FieldSchema} = Field | Fields], Conf0, Acc, Opts) ->
         end,
     Conf1 = put_value(Opts, FieldName, unbox(Opts, FValue), Conf0),
     %% now drop all aliases
+    %% it is allowed to have both old and new names provided in the config
+    %% but the first match wins.
     Conf = maps:without(Aliases, Conf1),
     map_fields(Fields, Conf, FAcc ++ Acc, Opts).
 
--- a/src/hocon_tconf.erl
+++ b/src/hocon_tconf.erl
@@ -294,13 +294,6 @@ map(Schema, Conf0, Roots0, Opts0) ->
     ),
     Conf1 = ensure_format(Conf0, Opts),
     Roots = resolve_root_types(hocon_schema:roots(Schema), Roots0),
-    %% assert
-    lists:foreach(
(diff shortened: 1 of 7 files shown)


lafirest (approved):

src/hocon_tconf.erl
```diff
@@ -478,7 +456,9 @@ map_fields_cont([{FieldName, FieldSchema} | Fields], Conf0, Acc, Opts) ->
                 ),
                 erlang:raise(C, Err, St)
         end,
-    Conf = put_value(Opts, FieldName, unbox(Opts, FValue), Conf0),
+    Conf1 = put_value(Opts, FieldName, unbox(Opts, FValue), Conf0),
+    %% now drop all aliases
+    Conf = maps:without(Aliases, Conf1),
```
keynslug, line 461: I personally feel that I'm missing some semantics: does it mean that either the field name itself or the _first_ defined alias for a field will win in a conf file? Maybe worth to clarify it in [this comment](https://github.com/emqx/hocon/pull/235/files#diff-1f8bb16d97e0cfdeb0298b5a722454f2ad949f4491ca49f8e4971da0b33b01b8R136-R139)?

src/hocon_tconf.erl
```diff
@@ -433,18 +410,19 @@ do_map2(Fields, Value0, Opts) ->
...
-            map_fields_cont([{FieldName, FieldSchema} | Fields], Conf0, Acc, Opts)
+            map_fields_cont([Field | Fields], Conf0, Acc, Opts)
     end.
 
-map_fields_cont([{FieldName, FieldSchema} | Fields], Conf0, Acc, Opts) ->
+map_fields_cont([{_, FieldSchema} = Field | Fields], Conf0, Acc, Opts) ->
     FieldType = field_schema(FieldSchema, type),
-    FieldValue = get_field(Opts, FieldName, Conf0),
+    [FieldName | Aliases] = name_and_aliases(Field),
+    FieldValue = get_field_value(Opts, [FieldName | Aliases], Conf0),
```
keynslug, line 425: Also, is it right that the schema generally allows alias clashes? I see that the `hocon_schema_json` deals with this, but here it seems to be allowed? Again, worth describing in the same comment what are the implications, e.g. what alias / field would win in a clash?

keynslug (commented):
Cool! 🔥

src/hocon_tconf.erl
zmstone: the duplication check can be done at compile time -- this is why it's moved to hocon_schema_json which is a compile-time dump.

it does not make sense to repeatedly check schema sanity at runtime.

keynslug (approved):

zmstone:
tagged 0.37.0

PR #237, opened 2023-03-17
fix(json): hide hidden fields from json schema dump

thalesmg:


src/hocon_schema.erl
```diff
@@ -298,6 +302,27 @@ maybe_add_desc(Mod, Name, Meta) ->
...
+inherit_hidden(Parent, Fields) when is_map(Fields) ->
+    case Parent of
+        #{hidden := true} ->
+            Fields#{hidden => true};
+        _ ->
+            Fields
+    end.
+
+maybe_add_hidden(Meta, Mod, Name) ->
+    case erlang:function_exported(Mod, hidden, 1) of
```
zmstone, line 314: we should not need to support this.
root schema can also be defined as map

src/hocon_schema_json.erl
```diff
@@ -49,9 +49,17 @@ gen(Schema, Opts) ->
...
-                fun({Ns, Name, Fields}) ->
-                    gen_struct(RootNs, Ns, Name, Fields, Opts2)
+            | lists:filtermap(
+                fun
+                    ({_Ns, _Name, _Meta = #{hidden := true}}) ->
+                        false;
+                    ({Ns, Name, Meta}) ->
+                        case gen_struct(RootNs, Ns, Name, Meta, Opts2) of
+                            #{fields := []} ->
+                                false;
```
zmstone, line 59: should rather raise an exception? so to mark the parent struct hidden.

src/hocon_schema.erl
```diff
@@ -531,6 +593,12 @@ is_deprecated(Schema) ->
     IsDeprecated = field_schema(Schema, deprecated),
     IsDeprecated =/= undefined andalso IsDeprecated =/= false.
 
+%% @doc Return 'true' if the field is marked as hidden.
+-spec is_hidden(field_schema()) -> boolean().
+is_hidden(Schema) ->
+    IsHidden = field_schema(Schema, hidden),
+    IsHidden =/= undefined andalso IsHidden =/= false.
```
zmstone, line 600: ?
```suggestion
    IsHidden =:= true.
```

src/hocon_schema.erl
```diff
@@ -360,10 +385,47 @@ find_ref(Schema, Name, Acc, Stack, TStack) ->
             Acc;
         false ->
             Fields0 = fields_and_meta(Schema, Name),
```
zmstone, line 387: can we simply drop the hidden fields from here so there is no need to recursively resolve the children fields ?
i.e. do not expose the `hidden or not` state to `hocon_schema_json` and `hocon_schema_example`.

but we need to switch though, so we can choose to either dump a minimal schema (including hidden fields), or a full schema (including hidden fields)

zmstone: also filter `?MODULE:roots/2` result too.
we'll maybe have to introduce a `Opt` or process-dictionary flag, to set `ignore_hidden` to `true`, then drop hidden roots and struct-fields if the flag is set.

thalesmg: >  or process-dictionary flag

I confess this was the first thing that came to mind, given the mutually recursive calls of `find_structs` :sweat_smile:

src/hocon_schema.erl
thalesmg: Thanks for the explanation, remove this function.

src/hocon_schema_json.erl
```diff
@@ -50,8 +50,22 @@ gen(Schema, Opts) ->
...
+                    case gen_struct(RootNs, Ns, Name, Meta, Opts2) of
+                        #{fields := []} = Meta1 ->
+                            error(
+                                {struct_with_no_fields, #{
+                                    namespace => Ns,
+                                    name => Name,
+                                    meta => Meta1,
+                                    msg =>
+                                        "If all children are hidden fields,"
+                                        " please set the parent struct as hidden."
```
zmstone, line 63:
```suggestion
                                        " please set the parent field as hidden."
```

src/hocon_schema.erl
```diff
@@ -360,17 +380,47 @@ find_ref(Schema, Name, Acc, Stack, TStack) ->
...
-            Fields = Fields0#{paths => Paths#{Path => true}},
+            Paths = Paths0#{Path => true},
+            Fields1 =
+                case get_include_hidden_fields() of
+                    true ->
+                        Fields0;
+                    false ->
+                        reject_hidden(Fields0)
```
zmstone, line 389: nit:
```suggestion
                        drop_hidden(Fields0)
```

src/hocon_schema.erl
```diff
@@ -298,6 +303,14 @@ maybe_add_desc(Mod, Name, Meta) ->
             Meta
     end.
 
+inherit_hidden(Parent, Fields) when is_map(Fields) ->
```
zmstone, line 306: do we still need this?

zmstone (commented):
great job.

src/hocon_schema.erl
thalesmg: I thought this was needed, but later I added the `reject_hidden` filter it seems no longer necessary.

src/hocon_schema.erl
```diff
@@ -360,17 +370,47 @@ find_ref(Schema, Name, Acc, Stack, TStack) ->
...
+                        Fields0;
+                    false ->
+                        reject_hidden(Fields0)
+                end,
+            Fields = Fields1#{paths => Paths},
             find_structs(Schema, Fields, Acc#{Key => Fields}, Stack, [Key | TStack])
     end.
 
-%% @doc Collect all structs defined in the given schema.
+reject_hidden(Meta = #{fields := _}) ->
```
thalesmg, line 385:
```suggestion
drop_hidden(Meta = #{fields := _}) ->
```

zmstone (approved):

PR #238, opened 2023-03-22
0322 do not fill default value if hidden

zmstone:
When a field is `hidden` it would not make sense to fill default value for it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,5 @@ generated/
 
 # git shallow clone does not work with rebar.lock
 rebar.lock
+
+erlang_ls.config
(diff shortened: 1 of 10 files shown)


erlang_ls.config
thalesmg, line 1: Maybe this shouldn't be commited?

Erlang LSP works locally without this config, and the paths given here are all pointing to the build directory rather than the source dir.

thalesmg (dismissed):

thalesmg (approved):

thalesmg (dismissed):

thalesmg (approved):

PR #240, opened 2023-03-23
feat: return #{gateway => #{}} when gateway = {}

zhongwencool:
fix：
```
v5.0.21-gddffba03(emqx@127.0.0.1)56> hocon:binary(<<“gateway={}“>>).
{ok,#{}}
```
after:
```
v5.0.21-gddffba03(emqx@127.0.0.1)56> hocon:binary(<<“gateway={}“>>).
{ok,#{<<"gateway">> => #{}}}
```

### DIFF
--- a/src/hocon.erl
+++ b/src/hocon.erl
@@ -320,7 +320,7 @@ do_concat(Concat, Location) ->
     do_concat(Concat, Location, []).
 
 do_concat([], _, []) ->
-    nothing;
+    #{?HOCON_T => object, ?HOCON_V => []};
 do_concat([], MetaKey, [{#{?METADATA := MetaFirstElem}, _V} = F | _Fs] = Acc) when ?IS_FIELD(F) ->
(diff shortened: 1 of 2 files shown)


zhongwencool (commented):
Support `null` to do replacement action, looks great! 
There should be this case in the future.
But I find 
```
hocon:binary("a={b = 1},a=whatever-string,a = {}").
{ok,#{<<"a">> => #{}}}
```
Not only 'null' can do replacement, any string will have the same effect as 'null', Is this also the expected behavior?

zmstone:
> Support `null` to do replacement action, looks great! There should be this case in the future. But I find
> 
> ```
> hocon:binary("a={b = 1},a=whatever-string,a = {}").
> {ok,#{<<"a">> => #{}}}
> ```
> 
> Not only 'null' can do replacement, any string will have the same effect as 'null', Is this also the expected behavior?

Yes, any value should do, except for objects.

PR #244, opened 2023-04-06
refactor: drop support for internal i18n resolution

zmstone:
refactor: drop support for internal i18n resolution

    From now on type description (doc) multi-lang resultion
    has to be done in a injected desc resolution callback.

    Or provide no callback and use the dummy resolver
    which generates a reference to the description
    so it can be resolved later.


thalesmg (approved):
So, in EMQX's side, the doc generation step will have to provide something like `desc_resolved => fun(Desc) -> read(LangFile, Desc) end`?

thalesmg (approved):

PR #249, opened 2023-04-24
fix: detect bad escape sequence

zmstone:
Do not allow unescaped quotes between `"..."`

`key="a"b"` is a syntax error, the valid expression should be `key="a\"b"`
-- this is as expectedd.

However, prior to this change, if we put two backslashes before the `"`, it becomes valid.
i.e. below two files are parsed to the same result:
```
key="a\\"b"
```
and
```
key="a\\\"b"
```

after the fix, `key="a\\"b"` becomes invalid.


zhongwencool (approved):

zmstone:
Fixing this may cause some compatibility issue.

zmstone:
https://github.com/emqx/hocon/pull/250

PR #259, opened 2023-06-16
fix: builtin type don't convert from_string when value is integer

zhongwencool:
When an integer value is read from a hocon file or default value, `typerelf:from_string` conversion is not executed, resulting in an unchecked boundary condition.
More detail see hocon_schema_builtin_tests.erl's `to_ip_port/1` function.


kjellwinblad (approved):

kjellwinblad (approved):
I'm not a hocon expert but it looks good to me.

PR #263, opened 2023-07-03
fix: make sure we don't accept fractional values

sstrigler:


src/hocon_postprocess.erl
```diff
@@ -152,7 +152,7 @@ bytesize(Other) ->
     Other.
 
 do_bytesize(Str) ->
-    {ok, MP} = re:compile("([0-9]+)(kb|KB|mb|MB|gb|GB)$"),
+    {ok, MP} = re:compile("^\s*([0-9]+)(kb|KB|mb|MB|gb|GB)$"),
```
paulozulato, line 155: should we accept spaces in the beginning of the string (but not in the end neither between then number and the unit)?

paulozulato (approved):

src/hocon_postprocess.erl
sstrigler: Was wondering the same but didn't want to change too much with this quick PR. I think the original intention for why there was no "^" was to allow spaces at the beginning (which might stem from dashboard inputs, the unit is added by the dashboard and thus no trailing spaces can occur).

PR #266, opened 2023-07-26
fix: updates return type for `desc` callback

paulozulato:
Fix return type for `desc` callback

### DIFF
--- a/src/hocon_schema.erl
+++ b/src/hocon_schema.erl
@@ -67,7 +67,7 @@
 -callback translations() -> [name()].
 -callback translation(name()) -> [translation()].
 -callback validations() -> [validation()].
--callback desc(name()) -> iodata() | undefined.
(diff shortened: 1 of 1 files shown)


src/hocon_schema.erl
```diff
@@ -67,7 +67,7 @@
 -callback translations() -> [name()].
 -callback translation(name()) -> [translation()].
 -callback validations() -> [validation()].
--callback desc(name()) -> iodata() | undefined.
+-callback desc(name()) -> desc().
```
thalesmg, line 70: I think there are several usages in EMQX that return `undefined` as well. :thinking: 

```suggestion
-callback desc(name()) -> desc() | undefined.
```

paulozulato: Ops... fixed (curiously dialyzer didn't complain about it locally...)

thalesmg (approved):

PR #267, opened 2023-08-04
feat: add unsafe atom conversion to `hocon_tconf:check_plain`

thalesmg:
By using `hocon_tconf:check_plain(Schema, Map, #{atom_key => {true, unsafe}})`, fields and `hoconsc:map` keys are converted to atoms unsafely, while keys in `map()` and their nested values therein are kept as binaries.


thalesmg:
tagged `0.39.16`

PR #282, opened 2023-12-13
feat: pp supports wrapped value

lafirest:


zmstone (approved):

keynslug:
BTW, `emqx_schema_secret` implements [`make_serializable`](https://github.com/emqx/emqx/blob/abeb5e985f48131bc97d8208f9ce7867cacb75af/apps/emqx/src/emqx_schema_secret.erl#L56-L57), I wonder if it could have helped instead.

PR #290, opened 2024-03-03
breaking: no escape at all in triple quotes

zmstone:
Prior to this breaking change, backslashes inside triple-quotes are treated as escape sequence.
This design allowed the last string character to be `"` as long as it's escaped.

```
key = """\"string value with quotes\""""
```
Which had its advantages however lead to the need of escaping backslash itself.

The need of escaping backslash inside triple-quote string
not only violate the HOCON spec, but also make it hard for users
to hand-craft multe-line string values.

For example, when writing EMQX rule-engine SQL program:

```
sql = """SELECT tokens(payload, b'\n')"""
```
would actually require users to write

```
sql = """SELECT tokens(payload, b'\\n')"""
```
 -- which is quite counter counterintuitive.

The cost of having to escape `\` will probably be higher
than the gain of supporting escape sequence in tripe-quote strings.

After this change, 4-quotes is syntax error, and tripe-quote as value is no longer possible (due to lack of escape sequence). 

However, there are work arounds:

-  A string ending with `"` can be wrapped with `"""~` and `~"""` instead.

    e.g.
    ```
    key = """~
        "string value with quotes"~"""
    ```
 
- Use normal quotes with escape sequence: `"\"\"\""`
- Make use of string concatenation feature: quote the proceeding characters using tripe-quotes, quote `"""` as `"\"\"\""`, then quote the following characters with triple-quotes again.
    e.g.
    ```
    key = """~
        line1
    ~"""
    "line2\"\"\"\n"
    """~
        line3
    ~"""
    ```


zhongwencool (approved):
👍

PR #294, opened 2024-05-21
Sync from feat/customizable_key

lafirest:


test/hocon_tconf_tests.erl
```diff
@@ -2466,8 +2505,7 @@ map_atom_keys_test_() ->
                 ?assertThrow(
                     {_, [
                         #{
-                            kind := validation_error,
```
zmstone, line 2469: why is this `kind` field dropped? 
emqx code for example, may catch and parse/format the errors list for logging, theses formatters may expect errors to have a `kind`.

zmstone (approved):

PR #297, opened 2024-07-03
fix: re:run does not work with unicode char list

zmstone:


thalesmg (approved):

zmstone:
thanks for the review.
 force pushed tag 0.43.1

src/hocon_tconf.erl
```diff
@@ -1352,6 +1352,9 @@ default_map_key_name_validator(Name) ->
             end
     end.
 
+re_run(Str, Pattern) ->
+    re:run(unicode:characters_to_binary(Str, utf8), Pattern).
```
keynslug, line 1356: Nit. I think the [correct way](https://www.erlang.org/doc/apps/stdlib/re.html#t:compile_option/0) is actually `re:run(Str, Pattern, [unicode])`. Otherwise the risk is `re` consuming half of the (encoded) codepoint somewhere, because it still thinks that one byte = one character. Should be okay with current `?MAP_KEY_RE` but not too future-proof.

PR #307, opened 2025-04-22
fix: prevent trailing backslash from "eating" triple quotes

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14157

### DIFF
--- a/src/hocon_scanner.xrl
+++ b/src/hocon_scanner.xrl
@@ -50,11 +50,14 @@ Float               = {Integer}?{Fraction}|{Integer}{Fraction}{Exponent}
 
 %% String
 Hex                 = [0-9A-Fa-f]
-Escape              = ["\\bfnrt]
+EscapeNoQuote       = [\\bfnrt]
+Escape              = "|{EscapeNoQuote}
 UnicodeEscape       = u{Hex}{Hex}{Hex}{Hex}
 Char                = ([^\"{LineFeed}]|\\{Escape}|\\{UnicodeEscape})
 String              = "{Char}*"
-MultilineChar       = ([^"]|"[^"]|""[^"]|\\{Escape}|\\{UnicodeEscape})
+%% Special handling for trailing quote: if we don't assert it's not followed by two other
+%% quotes, `{Escape}` would "eat" one of the quotes in the triple quote...
+MultilineChar       = (\\"[^"][^"]|[^"]|"[^"]|""[^"]|\\{EscapeNoQuote}|\\{UnicodeEscape})
 MultilineString     = """{MultilineChar}*"""
 
 %% Bytesize and Duration
--- a/src/hocon_scanner.xrl
+++ b/src/hocon_scanner.xrl
@@ -57,7 +57,7 @@ Char                = ([^\"{LineFeed}]|\\{Escape}|\\{UnicodeEscape})
 String              = "{Char}*"
 %% Special handling for trailing quote: if we don't assert it's not followed by two other
 %% quotes, `{Escape}` would "eat" one of the quotes in the triple quote...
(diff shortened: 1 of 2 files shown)


src/hocon_scanner.xrl
```diff
@@ -50,11 +50,14 @@ Float               = {Integer}?{Fraction}|{Integer}{Fraction}{Exponent}
...
 Hex                 = [0-9A-Fa-f]
-Escape              = ["\\bfnrt]
+EscapeNoQuote       = [\\bfnrt]
+Escape              = "|{EscapeNoQuote}
 UnicodeEscape       = u{Hex}{Hex}{Hex}{Hex}
 Char                = ([^\"{LineFeed}]|\\{Escape}|\\{UnicodeEscape})
 String              = "{Char}*"
-MultilineChar       = ([^"]|"[^"]|""[^"]|\\{Escape}|\\{UnicodeEscape})
+%% Special handling for trailing quote: if we don't assert it's not followed by two other
+%% quotes, `{Escape}` would "eat" one of the quotes in the triple quote...
```
zmstone, line 59: could you help to explain why this "eat" happens?

This should work?
```
([^"]|"[^"]|""[^"]|\\{EscapeNoQuote}|\\{UnicodeEscape})
```

src/hocon_scanner.xrl
```diff
@@ -50,11 +50,14 @@ Float               = {Integer}?{Fraction}|{Integer}{Fraction}{Exponent}
...
+MultilineChar       = (\\"[^"][^"]|[^"]|"[^"]|""[^"]|\\{EscapeNoQuote}|\\{UnicodeEscape})
```
zmstone, line 60: `\\{Escap}` is changed to `\\{EscapenoQuote}`
does it mean `\\` is now parsed as `\\`, but not `\` ?

zmstone: nvm, \\ is in EscapNoQuote, it's no quote, not no backslash.


I wonder if \\ should have been added in MultilineChar at all.
it's a breaking change if we remove it though.

src/hocon_scanner.xrl
thalesmg: The eat happens because the dangling backslash is the last thing the triple quoted string.  `...\\"""` essentially eats the last quote and breaks out of the triple quote.

```
"""aaa\"""\nx = 10\ny = """~...
```

Since `\` was not being escaped when "unparsing" the config, it would eat the triple quote, and everything after that gets interpreted as the string until it hits another triple quote.

It's easy to see if you try the test case out without the code changes.

src/hocon_scanner.xrl
thalesmg: not sure I understand the question.

the regex for `EscapeNoQuote` is simple the old `Escape` without `"`.

thalesmg: I've just tried your suggestion, and with the fix I added to escape contents of triple quotes that are not multiline, it seems to work.

src/hocon_scanner.xrl
thalesmg: I thought I had tried your suggestion before, but paired with other changes (escaping or not escaping the string, various combinations of changes), but didn't work before, but now the added test case works with that regex.  :thinking:

zmstone (approved):

thalesmg:
tagged 0.45.3

src/hocon_scanner.xrl
thalesmg: P.S.: somehow, GH showed only your 1st comment when I wrote my 1st reply :see_no_evil: :hear_no_evil: :speak_no_evil:

PR #314, opened 2026-02-08
fix: integer enum format error

zhongwencool:
https://github.com/emqx/emqx/pull/16719


zmstone:
why not integer type with a range validator ?

zmstone:
enum symbols are atoms btw.

PR #317, opened 2026-07-31
fix: process unicode escape sequences

keynslug:
## Summary

This PR adds missing processing for JSON-compatible Unicode escape sequences, adapted from OTP-28's `json` module.


src/hocon_scanner.xrl
```diff
@@ -232,8 +232,55 @@ unesc([$\\, $t | T]) -> {$\t, T};
 unesc([$\\, $r | T]) -> {$\r, T};
 unesc([$\\, $b | T]) -> {$\b, T};
 unesc([$\\, $f | T]) -> {$\f, T};
+unesc([$\\, $u | T]) -> unescapeu(T);
```
zmstone, line 235: this only unescapes double quote strings.
for """ strings, maybe add a test case to cover that it's intentionally not unescaped

zmstone (approved):

PR #318, opened 2026-08-04
fix(cli): stage generated config files before commit

yuanweize:
## Problem

When disk space is exhausted (`ENOSPC`), writing generated `app.<time>.config` and `vm.<time>.args` directly can create or truncate final files before the write fails. A later startup may then pick up a 0-byte or incomplete generated configuration.

Failure-path review also found that earlier revisions could delete an existing target, publish only one of the two generated files, accumulate abandoned temp files, prune the just-written generation, or fail to reclaim excess history before an ENOSPC retry.

## Solution

- Stage both generated outputs in unique, exclusive temporary files in the destination directory.
- Apply an existing regular target's permission bits to its staging/rollback file before writing content.
- Write, `sync`, and close every temporary file before committing either output.
- Keep a synced rollback copy only when replacing a pre-existing target.
- Rename `vm.args` first and `app.config` last.
- If a later rename returns an error, roll back earlier commits in reverse order:
  - remove a target created by this generation; or
  - restore the previous target contents from its rollback copy.
- Clean up uncommitted temporary files and rollback copies on handled failure paths.
- Remove stale temp files only when they match the exact generated target family and numeric temp suffix format.
- Prune excess history before staging to reclaim space, excluding the current destination and reserving one slot for it; prune again after commit to enforce the final limit.
- Match history by the complete destination base, extension, and timestamp shape so a dotted `--dest_file` cannot prune another config family.

This provides rollback for handled staging/rename errors. It does not claim a strict two-file ACID transaction across process crashes, concurrent generators, or power loss.

## Regression tests

Coverage includes:

- staging and backup-preparation failures preserving existing targets;
- an injected second-rename failure removing a partial new generation;
- the same failure restoring pre-existing target contents;
- an older explicit timestamp retaining the current generation during pruning;
- exact file-family pruning for dotted destination names;
- stale temp cleanup without deleting fresh, unrelated-family, or non-generated temp files;
- replacement preserving existing regular-file permission bits;
- a failed backup restore preserving the current target and logging the rename error.

## Testing

- Erlang/OTP 23: production `rebar3 compile` passed
- Erlang/OTP 26: Elvis passed; both changed files pass erlfmt
- Erlang/OTP 27: Dialyzer, xref, full EUnit, and coverage passed — 584 tests
- Erlang/OTP 28: Dialyzer, full EUnit, and coverage passed — 584 tests
- Erlang/OTP 28: `hocon_cli_tests` passed — 25 tests
- Real 1 MiB tmpfs ENOSPC injection:
  - second staging write failed with both old targets intact and no temp files;
  - rollback-backup preparation failed with both old targets intact and no temp files;
  - pre-prune reclaimed enough space for generation to succeed with the final retention limit enforced.
- `git diff --check` passed

### DIFF
--- a/src/hocon_cli.erl
+++ b/src/hocon_cli.erl
@@ -18,11 +18,15 @@
 
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
--export([is_valid_now_time/1]).
+-export([atomic_write_files/2, is_valid_now_time/1, restore_backup/2]).
 -endif.
 
 -export([main/1]).
 
+-include_lib("kernel/include/file.hrl").
+
+-define(NOW_TIME_LENGTH, 19).
+-define(STALE_TMP_AGE_SECONDS, 3600).
 -define(STDERR(Str, Args), io:format(standard_error, Str ++ "~n", Args)).
 -define(STDOUT(Str, Args), io:format(Str ++ "~n", Args)).
 -define(FORMAT_TEMPLATE, [time, " [", level, "] ", msg, "\n"]).
@@ -306,6 +310,9 @@ generate(ParsedArgs) ->
     DestinationVMArgs = filename:join(AbsPath, DestinationVMArgsFilename),
     log(debug, "Generating config in: ~0p", [Destination]),
 
+    cleanup_stale_tmp_files(Destination),
+    cleanup_stale_tmp_files(DestinationVMArgs),
+
     Schema = load_schema(ParsedArgs),
     Conf = load_conf(ParsedArgs, fun log/3),
     LogFun =
@@ -319,22 +326,21 @@ generate(ParsedArgs) ->
             AppConfig = proplists:delete(vm_args, NewConfig),
             VmArgs = stringify(proplists:get_value(vm_args, NewConfig)),
 
-            %% Prune excess files
-            MaxHistory = proplists:get_value(max_history, ParsedArgs, 3) - 1,
+            Files = [
+                {DestinationVMArgs, string:join(VmArgs, "\n")},
+                {Destination, io_lib:fwrite("~p.\n", [AppConfig])}
+            ],
+            MaxHistory = proplists:get_value(max_history, ParsedArgs, 3),
+            %% Reclaim excess history before staging so an ENOSPC retry can recover.
             prune(Destination, MaxHistory),
             prune(DestinationVMArgs, MaxHistory),
-
-            case
-                {
-                    file:write_file(Destination, io_lib:fwrite("~p.\n", [AppConfig])),
-                    file:write_file(DestinationVMArgs, string:join(VmArgs, "\n"))
-                }
-            of
-                {ok, ok} ->
-                    ok;
-                {Err1, Err2} ->
-                    maybe_log_file_error(Destination, Err1),
-                    maybe_log_file_error(DestinationVMArgs, Err2),
+            case atomic_write_files(Files) of
+                ok ->
+                    %% Enforce the retention limit after both files are in place.
+                    prune(Destination, MaxHistory),
+                    prune(DestinationVMArgs, MaxHistory);
+                {error, Filename, Reason} ->
+                    log(error, "Error writing ~s: ~s", [Filename, file:format_error(Reason)]),
                     stop_deactivate()
             end
     catch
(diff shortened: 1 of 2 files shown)


yuanweize:
Follow-up commit `f975484` addresses the review feedback and hardens the full generation failure path:

- final targets are never deleted on failure;
- both outputs are staged and synced before commit;
- history is pruned only after successful commit;
- temporary filenames are unique;
- the regression test now verifies that a failed generation preserves old app/vm history and publishes no new app config.

Local OTP 27/28 xref + full EUnit pass (563 tests each), OTP 27 Dialyzer passes, and OTP 26 style checks pass.

The three GitHub Actions runs are currently marked `action_required`, so they appear to need maintainer approval before they can execute. Please re-review when convenient.

src/hocon_cli.erl
```diff
@@ -390,12 +386,79 @@ do_delete([File | Files], Left) ->
...
     ok;
-maybe_log_file_error(Filename, {error, Reason}) ->
-    log(error, "Error writing ~s: ~s", [Filename, file:format_error(Reason)]),
-    ok.
+commit_staged_files([{Filename, TmpFilename} | StagedFiles]) ->
+    case file:rename(TmpFilename, Filename) of
+        ok ->
+            commit_staged_files(StagedFiles);
+        {error, Reason} ->
+            cleanup_staged_files([{Filename, TmpFilename} | StagedFiles]),
```
zmstone: we write app config first, then vm args.
if vm args rename failed, the already renamed app config file stays.
this breaks the "all or nothing" ambition of this pr

yuanweize: You are right about the partial-commit issue. The current branch commits `vm.args` first and `app.config` last, but the failure is symmetric: the first successful rename could remain if the second rename failed.

Commit `cbb7088` now keeps synced rollback copies for pre-existing targets and tracks committed files. If a later rename returns an error, it rolls back earlier commits in reverse order: newly created targets are removed, while replaced targets are restored to their previous contents. Uncommitted temp/backup files are also cleaned up.

I added injected second-rename tests for both cases (targets initially absent and targets already existing). OTP 27/28 full EUnit passes with 565 tests; xref, Dialyzer, erlfmt, and Elvis also pass.

src/hocon_cli.erl
```diff
@@ -319,22 +319,18 @@ generate(ParsedArgs) ->
...
-                    maybe_log_file_error(DestinationVMArgs, Err2),
+            Files = [
+                {DestinationVMArgs, string:join(VmArgs, "\n")},
+                {Destination, io_lib:fwrite("~p.\n", [AppConfig])}
+            ],
+            case atomic_write_files(Files) of
+                ok ->
+                    %% Prune only after both new files are safely in place.
+                    MaxHistory = proplists:get_value(max_history, ParsedArgs, 3),
+                    prune(Destination, MaxHistory),
```
zmstone, line 340: Pruning after the write now includes the just-written file in the candidate set. `prune/2` keeps the `MaxHistory` lexicographically-largest names, so if the new generation's timestamp sorts older than the existing files (a `-t` value older than existing generations, or the wall clock stepping backwards between runs), the file we just committed is the one that gets deleted — `generate` still reports success while the newest-by-name config on disk is a stale one. The old code couldn't hit this because it never pruned the file it was about to write.

Suggest excluding `Destination`/`DestinationVMArgs` from the prune candidates.

src/hocon_cli.erl
```diff
@@ -390,12 +386,154 @@ do_delete([File | Files], Left) ->
...
+            ({_Filename, _TmpFilename, BackupFilename}) -> _ = file:delete(BackupFilename)
+        end,
+        PreparedFiles
+    ).
+
+temporary_filename(Filename) ->
+    lists:flatten(
+        io_lib:format(
+            "~s.~s.~B.tmp",
+            [Filename, os:getpid(), erlang:unique_integer([positive, monotonic])]
```
zmstone, line 647: If the escript dies between staging and cleanup (e.g. kill -9), these `.tmp` files are left behind and nothing ever removes them: `prune/2`'s wildcard is `Base ++ ".*" ++ Ext` with `Ext` being `.config`/`.args`, so `*.tmp` never matches, and every later run generates fresh unique names. They accumulate in the config dir forever.

Suggest cleaning up stale `app.*.config.*.tmp` / `vm.*.args.*.tmp` leftovers at the start of `generate`, or including them in prune.

src/hocon_cli.erl
```diff
@@ -390,12 +386,154 @@ do_delete([File | Files], Left) ->
...
+rollback_committed_file({Filename, _TmpFilename, BackupFilename}) ->
+    restore_backup(Filename, BackupFilename).
+
+restore_backup(Filename, BackupFilename) ->
+    case file:rename(BackupFilename, Filename) of
+        ok ->
+            ok;
+        {error, _Reason} ->
+            _ = file:delete(Filename),
+            maybe_log_rollback_error(Filename, file:rename(BackupFilename, Filename))
```
zmstone: just log the error.
file:rename/1 atomically rename and overwrite existing.
attempting to delete will fail wit the same error, and retry won't help.

yuanweize:
Follow-up commit `29d6e88` addresses all three latest maintainer review findings:

- **Pruning:** the just-written generation is explicitly kept; only `max_history - 1` older files are eligible for deletion, including the `max_history=0` boundary.
- **Stale staging files:** old `app.*.config.*.tmp` / `vm.*.args.*.tmp` files are cleaned at generation start when older than one hour. Fresh temp files are left alone to avoid deleting another active generator's staging file.
- **Rollback:** failed backup restoration now only logs the rename error; it no longer deletes the target or retries a filesystem operation that already failed.

Added regression coverage for older timestamps, stale temp cleanup, and rollback failure preservation. Verified in clean OTP 27 and OTP 28 containers: 22 `hocon_cli_tests` passed on each; OTP 28 erlfmt check passed; `git diff --check` passed.

Please re-review the latest commit when convenient.

yuanweize:
The branch was updated to `2883fc7` (force-with-lease from the previously pushed `29d6e88`) to add a direct regression test for the rollback error path. The implementation remains unchanged: failed backup restoration only logs the single failed rename and does not delete or retry the target.

Final clean OTP 28 verification: `rebar3 fmt --check src/hocon_cli.erl` passed; `rebar3 as test eunit --module=hocon_cli_tests` passed (22 tests, 0 failures). OTP 27 had the same 22/0 result.

Please review the latest head `2883fc7`.

src/hocon_cli.erl
yuanweize: Good catch. Fixed in `2883fc7`: `prune/2` now removes the current output from the candidate list and keeps at most `MaxHistory - 1` older files. I added a regression test where `-t` is older than the existing generations.

src/hocon_cli.erl
yuanweize: Fixed in `2883fc7`. Generation now removes matching temp files older than one hour, while leaving recent ones alone to avoid racing another generator. The test covers both stale and fresh temp files.

src/hocon_cli.erl
yuanweize: Agreed. The delete/retry path was removed in `2883fc7`. Follow-up `1939178` also separates delete cleanup from restore failures, so an `enoent` from the backup rename is logged instead of being silently treated like an already-deleted new target. The regression test leaves the current target untouched and now exercises the logged error path.

src/hocon_cli.erl
```diff
@@ -319,22 +325,18 @@ generate(ParsedArgs) ->
...
+            Files = [
+                {DestinationVMArgs, string:join(VmArgs, "\n")},
+                {Destination, io_lib:fwrite("~p.\n", [AppConfig])}
+            ],
+            case atomic_write_files(Files) of
+                ok ->
+                    %% Prune only after both new files are safely in place.
+                    MaxHistory = proplists:get_value(max_history, ParsedArgs, 3),
+                    prune(Destination, MaxHistory),
+                    prune(DestinationVMArgs, MaxHistory);
```
zmstone, line 341: With prune reachable only on success, the disk-full scenario this PR targets has lost its self-healing: previously prune ran *before* the write, so deleting old generations could free enough space for the new write to succeed. Now, on ENOSPC, staging fails first, prune never runs, and every retry fails identically until an operator manually deletes files. The staging pipeline also needs strictly more headroom than the old direct write (both new tmp files, plus full backup copies of any pre-existing targets from `backup_file/1`), so the failure threshold is reached sooner.

Suggest pruning history (excluding the current output, keeping `MaxHistory - 1`, i.e. exactly what `prune/2` now does) *before* staging as well — that only removes files already beyond the retention limit and can't touch the file about to be written — and keeping the post-commit prune as is.

src/hocon_cli.erl
```diff
@@ -369,12 +371,54 @@ prune(Filename, MaxHistory) ->
...
+    CurrentFilename = filename:basename(Filename),
     Files =
         lists:sort(filelib:wildcard(Base ++ ".*" ++ Ext, Path)),
+    HistoryFiles =
+        [filename:join([Path, F]) || F <- Files, F =/= CurrentFilename],
+    HistoryLimit =
+        case MaxHistory > 0 of
+            true -> MaxHistory - 1;
+            false -> 0
+        end,
```
zmstone: Nit: this is `HistoryLimit = max(MaxHistory - 1, 0)`.

zmstone (commented):
thank you, last two comments.

yuanweize:
Addressed the latest review feedback in `c01c610` (rebased onto current `master` `97b0f22`).

- Prune excess history before staging, so a disk-full retry can reclaim space; the current destination remains excluded.
- Keep the post-commit prune to enforce the final retention limit.
- Simplified `HistoryLimit` to `max(MaxHistory - 1, 0)`.
- Added/updated regression coverage for failed generation with retained history.
- Fixed the Style Check failure by removing duplicated test setup.

Validation after rebase:
- OTP 26 Elvis: passed
- OTP 26 erlfmt checks for changed files: passed
- OTP 27 xref + Dialyzer + 22 `hocon_cli_tests`: passed
- OTP 28 xref + full EUnit: 581 tests passed
- `git diff --check`: passed

yuanweize:
I found and fixed two additional filesystem edge cases in `1302f5f`: pruning for a dotted `--dest_file` could match another config family, and replacing an existing target changed its permission bits (for example, `0600` to the process default). Stale-temp cleanup is now also restricted to the exact generated family and numeric temp suffix format.

The new tests cover those cases plus second-target preparation failure. I reran clean OTP 23/26/27/28 checks and three real 1 MiB tmpfs ENOSPC scenarios; OTP 27 and 28 full EUnit both pass with 584 tests. The PR description has the exact current flow and validation results.

zmstone:
please rebase (and squash)

yuanweize:
Rebased onto latest `upstream/master` (`d2ffcc9`) and squashed the PR into a single commit (`936e64a`), as requested.

Verification after the rewrite:
- the PR patch is byte-for-byte identical to the pre-rebase patch;
- `git diff --check` passes;
- OTP 28 `hocon_cli_tests` passes: 25 tests, 0 failures.

No GitHub checks are reported yet for the rewritten head.

zmstone (approved):

zmstone:
thank you for the pr. 
tagged 0.46.2

PR #320, opened 2026-08-10
feat: redact sensitive values

thalesmg:


src/hocon_tconf.erl
```diff
@@ -533,6 +535,20 @@ map_one_field_non_hidden(FieldType, FieldSchema, FieldValue0, Opts) ->
             {Acc, FieldValue}
     end.
 
+maybe_redact_sensitive(FieldSchema, CheckedValue, #{
+    make_serializable := true, redact_sensitive := true
+}) ->
```
zmstone:
```suggestion
maybe_redact_sensitive(FieldSchema, CheckedValue, #{
    make_serializable := true, redact_sensitive := true
}) when CheckValue =/= undefined ->
```

src/hocon_tconf.erl
```diff
@@ -533,6 +535,20 @@ map_one_field_non_hidden(FieldType, FieldSchema, FieldValue0, Opts) ->
...
+    case field_schema(FieldSchema, sensitive) of
+        true ->
+            ?REDACTED_VAL;
+        {true, Fn} when is_function(Fn, 1) ->
+            Fn(CheckedValue);
```
zmstone: If Fn crashes, arg will be logged ?

src/hocon_tconf.erl
zmstone: there is already obfuscate/2.
but it needs to accept `{true, Fn}`, maybe extend it

src/hocon_tconf.erl
thalesmg: :checkered_flag:

src/hocon_tconf.erl
thalesmg: :checkered_flag:

zmstone (approved):

PR #323, opened 2026-08-28
feat: add hocon-convert CLI to emit JSON / YAML

keynslug:
## Summary

Add a separate simple CLI tool to produce JSON / YAML from HOCON files or HOCON stream read from standard input. Prompted by the recent EMQX Operator [change](https://github.com/emqx/emqx-operator/pull/1251), supposed to be used as auxiliary tool during migrations.

Tested manually. Additionally verified against several sample configs with `./hocon-convert --format yaml FILE | yq --output-format json`. Resulting JSON matched normalized direct `--format json` output:
 - test04.conf
 - test06.conf
 - test08.conf
 - test10.conf
 - test12.conf
 - variable-subst.conf
 - equiv02/path-keys.conf
 - equiv03/includes.conf


src/hocon_cli_convert.erl
```diff
@@ -0,0 +1,321 @@
...
+        },
+        notice
+    ),
+    logger:remove_handler(default),
+    logger:set_primary_config(level, LogLevel),
+    logger:add_handler(
+        hocon_cli,
+        logger_std_h,
+        #{
+            config => #{type => standard_io},
```
zmstone: The log handler writes to `standard_io`, which is also where the converted document goes with the default `--output -`. Any logger event (for example a VM error report) corrupts piped JSON/YAML output. Point the handler at `standard_error`.

Related: nothing in the `hocon:binary`/`hocon:files` parse path calls `logger` today, so `-l` currently has no observable effect.

src/hocon_cli_convert.erl
```diff
@@ -0,0 +1,321 @@
...
+            FD;
+        {error, Reason} ->
+            ?STDERR("Invalid output file: ~s", [file:format_error(Reason)]),
+            die(2)
+    end.
+
+write_output(FD, Bytes) ->
+    case write(FD, Bytes) of
+        ok ->
+            file:sync(FD);
```
zmstone: Two issues with the sync:
- `file:sync/1` accepts only a pid or `#file_descriptor{}`, so `file:sync(standard_io)` returns `{error, badarg}` on every stdout run.
- The result is discarded in both cases, so a real fsync failure on the `-o FILE` path (for example ENOSPC) is dropped and the tool exits 0.

Skip the sync for `standard_io` and check the result for file output.

src/hocon_cli_convert.erl
```diff
@@ -0,0 +1,321 @@
...
+format_conf(json, Conf) ->
+    json:format(Conf);
+format_conf(yaml, Conf) ->
+    format_yaml(Conf).
+
+slurp_standard_input() ->
+    slurp_standard_input([]).
+
+slurp_standard_input(Acc) ->
+    case io:get_chars("", 32768) of
```
zmstone: Nit: `io:get_chars/2` can also return `{error, Reason}`; that would crash with `case_clause` instead of taking the clean error path. Hard to hit on a pipe (current OTP silently falls back to latin1 on invalid UTF-8 instead of erroring — which also silently changes the parsed bytes), but a defensive `{error, Reason}` clause that reports and exits non-zero would be cleaner.

src/hocon_cli_convert.erl
```diff
@@ -0,0 +1,321 @@
...
+        false -> yaml_scalar(Key)
+    end;
+yaml_key(Key) ->
+    yaml_scalar(Key).
+
+is_plain_key(Key) ->
+    re:run(Key, <<"^[A-Za-z_][A-Za-z0-9_.-]*$">>, [{capture, none}]) =:= match.
+
+is_yaml_keyword(Key) ->
+    lists:member(string:lowercase(Key), [
```
zmstone: YAML 1.1 also treats `y`/`Y`/`n`/`N` as booleans, and `is_plain_key/1` accepts a single letter, so a key `y` is emitted unquoted. `gopkg.in/yaml.v2` loads `y: 1` as `{true: 1}` (checked with v2.4.0; PyYAML dropped the single-letter forms but still resolves `on`/`off`). Add `<<"y">>` and `<<"n">>` to the list — the lowercase pass already covers the uppercase forms. Only keys are affected; values always go through `json:encode`.

src/hocon_cli_convert.erl
```diff
@@ -0,0 +1,321 @@
...
+            "Specifies output format, `json` and `yaml` are supported"},
+        {log_level, $l, "log_level", {string, "notice"}, "Log level"}
+    ].
+
+main(Args) ->
+    case getopt:parse(cli_options(), Args) of
+        {ok, {Opts, PosArgs}} ->
+            case proplists:get_value(help, Opts) of
+                true ->
+                    usage(standard_error);
```
zmstone: Nit: explicitly requested help conventionally goes to stdout (`hocon-convert --help | less` shows nothing as-is). Keep `standard_error` for the invalid-args path and use `standard_io` here.

rebar.config
```diff
@@ -43,11 +43,22 @@
...
+    ]},
+    {cli_convert, [
+        {escript_name, hocon_cli_convert},
+        {escript_comment, "%% Hocon Convert 0.1.0"},
+        {escript_emu_args, "%%! +sbtu +A1 -escript main hocon_cli_convert\n"},
+        {post_hooks, [
+            {"(linux|darwin|solaris|freebsd|netbsd|openbsd)", escriptize,
+                "cp _build/cli_convert/bin/hocon_cli_convert ./hocon-convert"},
+            {"win32", escriptize,
+                "copy _build\\cli_convert\\bin\\hocon_cli_convert ./hocon-convert"}
```
zmstone: cmd.exe's built-in `copy` treats `/` anywhere in the argument line as a switch introducer, so `./hocon-convert` fails with "Invalid switch". Use `.\hocon-convert` (the sibling `cli` hook above uses plain `.`).

src/hocon_cli_convert.erl
```diff
@@ -0,0 +1,321 @@
...
+            filter_default => log,
+            filters => [],
+            level => all
+        }
+    ).
+
+prepare_output("-") ->
+    standard_io;
+prepare_output(Filename) ->
+    case file:open(Filename, [write, binary]) of
```
zmstone: should delay the open until all args/inputs are parsed.
otherwise it either truncates an existing good file, or leave behind an empty one.

scripts/elvis-check.sh
```diff
@@ -2,14 +2,25 @@
 
 set -euo pipefail
 
-ELVIS_VERSION='1.1.0-emqx-2'
+ELVIS_VERSION='3.2.6-emqx-1'
+ELVIS_SHA256='a9b92b72bc031cce3c6b743418e404a763303ca9f2bf3141eb8af8f22c4176fa'
 
 elvis_version="${2:-$ELVIS_VERSION}"
```
zmstone, line 8: there is no equivalent $3 for sha256, so the check at line 18 will be true, lead to always re-download.
maybe keep the checksum in a cache file too

src/hocon_cli_convert.erl
zmstone: explict request for usage print should be printed to stdio

src/hocon_cli_convert.erl
zmstone: logging to stdio will contaminate output (option `-`)?

src/hocon_cli_convert.erl
```diff
@@ -0,0 +1,321 @@
...
+        #{
+            "debug" => debug,
+            "info" => info,
+            "notice" => notice,
+            "warning" => warning,
+            "error" => error
+        },
+        notice
+    ),
+    logger:remove_handler(default),
```
zmstone: this module does not call logger
and  hocon:binary hocon:file don't call either.
so maybe there is no need

src/hocon_cli_convert.erl
```diff
@@ -0,0 +1,321 @@
...
+        {ok, FD} ->
+            FD;
+        {error, Reason} ->
+            ?STDERR("Invalid output file: ~s", [file:format_error(Reason)]),
+            die(2)
+    end.
+
+write_output(FD, Bytes) ->
+    case write(FD, Bytes) of
+        ok ->
```
zmstone:
```suggestion
        ok when FD =:= standard_io ->
            ok;
        ok ->
            ok = file:sync(FD);
          
```

src/hocon_cli_convert.erl
```diff
@@ -0,0 +1,321 @@
...
+    io:put_chars(standard_io, Bytes);
+write(FD, Bytes) ->
+    file:write(FD, Bytes).
+
+validate_format("json") -> json;
+validate_format("yaml") -> yaml;
+validate_format(Format) -> terminate_invalid_args({format, Format}).
+
+format_conf(json, Conf) ->
+    json:format(Conf);
```
zmstone: this is the first intro of json dependency to this repo.
need to check rebar.config and doc's claim of min Erlang/OTP version

src/hocon_cli_convert.erl
```diff
@@ -0,0 +1,321 @@
...
+format_conf(yaml, Conf) ->
+    format_yaml(Conf).
+
+slurp_standard_input() ->
+    slurp_standard_input([]).
+
+slurp_standard_input(Acc) ->
+    case io:get_chars("", 32768) of
+        eof ->
+            lists:reverse(Acc);
```
thalesmg: nano nit: seems like `hocon:binary` supports iolists, so we could just construct one that doesn't need reversing:

```elixir
:hocon.binary([[~c"a {" | "x = 1"], [], ?}])
{:ok, %{"a" => %{"x" => 1}}}
```

rebar.config
```diff
@@ -43,11 +43,22 @@
...
+    {cli, [
         {post_hooks, [
             {"(linux|darwin|solaris|freebsd|netbsd|openbsd)", escriptize,
-                "cp _build/es/bin/hocon ./"},
-            {"win32", escriptize, "copy _build\\es\\bin\\hocon ."}
+                "cp _build/cli/bin/hocon ./"},
+            {"win32", escriptize, "copy _build\\cli\\bin\\hocon ."}
+        ]}
+    ]},
+    {cli_convert, [
```
zmstone: wonder if we can merge the two script interfaces.

keep hocon_cli_convert.erl as lib.
cli entrypoint is 'hocon convert ...'

thalesmg (dismissed):

rebar.config
keynslug: The trouble is the main one is quite overloaded already.

There is stuff that looks pretty far out of scope of a HOCON library CLI (e.g. `now_time` subcommand, generated file archival and pruning), naming and handling of existing CLI options is slightly unconventional (e.g. no reading of stdin), and all existing subcommands share the same set of CLI options that I didn't want to contaminate further.

Anyway, I'll experiment what can be done about it.

src/hocon_cli_convert.erl
keynslug: > construct one that doesn't need reversing

You mean by using non-tail-recursive `slurp_standard_input/0`?

src/hocon_cli_convert.erl
keynslug: Right. Same with triple-quoted fixtures in unit tests. This is isolated to a separate CLI escript however, so my idea was to avoid bumping requirement of the whole library.

zmstone (dismissed):

rebar.config
keynslug: Drafted #324, appreciate an initial review.

PR #324, opened 2026-08-30
feat(cli): implement new unified CLI

keynslug:
## Summary

This PR add new unified CLI instead of `hocon_cli`. The original CLI is kept unchanged, so users calling it through `hocon_cli:main(...)` are unaffected.

This CLI:
* Incorporates most general functions from the original: `generate`, `check_schema`, `multi_get`, pretty-printing.
* Incorporates JSON/YAML formatting through `format` subcommand.
* Leaves stuff like `now_time`, filename generation, archival and pruning out of scope.

See also #323.


src/hocon_cli2.erl
```diff
@@ -0,0 +1,652 @@
...
+            case validate_arguments(ArgMap) of
+                ok ->
+                    Handler(ArgMap);
+                {error, Message} ->
+                    stderr("error: ~ts~n", [Message]),
+                    ?STATUS_USAGE_ERROR
+            end;
+        {error, Reason = {Path, _, _, _}} ->
+            stderr("error: ~ts~n", [argparse:format_error(Reason)]),
+            stderr("~ts", [argparse:help(Command, ParserOpts#{command => Path})]),
```
zmstone: tl(Path)

src/hocon_cli2.erl
```diff
@@ -0,0 +1,652 @@
...
+    ["--conf-file", Marker | mask_dash_arguments(Context, Marker, Rest)];
+mask_dash_arguments(Context, Marker, ["--conf-file=-" | Rest]) ->
+    ["--conf-file=" ++ Marker | mask_dash_arguments(Context, Marker, Rest)];
+mask_dash_arguments(format, Marker, ["-o", "-" | Rest]) ->
+    ["-o", Marker | mask_dash_arguments(format, Marker, Rest)];
+mask_dash_arguments(format, Marker, ["--output", "-" | Rest]) ->
+    ["--output", Marker | mask_dash_arguments(format, Marker, Rest)];
+mask_dash_arguments(format, Marker, ["--output=-" | Rest]) ->
+    ["--output=" ++ Marker | mask_dash_arguments(format, Marker, Rest)];
+mask_dash_arguments(format, Marker, ["-"]) ->
```
zmstone: [“-“ | Rest]
and continue on Rest

src/hocon_cli2.erl
```diff
@@ -0,0 +1,652 @@
...
+sync_logger() ->
+    try
+        logger_std_h:filesync(hocon_cli2)
+    catch
+        _:_ -> ok
+    end.
+
+add_code_paths(#{code_paths := Paths}) ->
+    lists:foreach(
+        fun(Path) -> true = code:add_patha(Path) end,
```
zmstone: handle error return from `code:add_patha/1`

src/hocon_cli2.erl
```diff
@@ -0,0 +1,652 @@
...
+        Paths
+    );
+add_code_paths(#{}) ->
+    ok.
+
+get_values(#{keys := Keys} = Parsed) ->
+    with_schema_and_conf(
+        Parsed,
+        fun(Schema, Conf) ->
+            RootNames = lists:usort([root_name(Schema, Key) || Key <- Keys]),
```
zmstone: Root resolution happens before the `try`, and `hocon_schema:resolve_struct_name/2` throws for an unknown root. Consequently, a typo such as `get ... doesnotexist.foo` escapes from `run/1` instead of producing one of the CLI status codes. Please validate or catch key resolution as part of the command error handling.

src/hocon_cli2.erl
```diff
@@ -0,0 +1,652 @@
...
+print_values(Values) ->
+    lists:foreach(fun({Key, Value}) -> io:format("~s=~0p~n", [Key, Value]) end, Values).
+
+write_generated_config(Args, Generated) ->
+    AppConfig = proplists:delete(vm_args, Generated),
+    AppOutput = maps:get(config_output, Args),
+    VMArgsOutput = maps:get(vm_args_output, Args),
+    AppContent = io_lib:fwrite("~p.~n", [AppConfig]),
+    VMArgs = stringify(proplists:get_value(vm_args, Generated)),
+    case write_output(AppOutput, AppContent) of
```
zmstone: These two outputs are written sequentially without a collision check or rollback. If both options name the same path, the command returns success but the VM-args write overwrites the app config; if the second write fails, the command returns `STATUS_OUTPUT_ERROR` while leaving the new app config behind. Please reject colliding destinations and commit the pair atomically (or roll back the first write).

zmstone:
we can maybe release a 1.0 drop the old cli?

now_time is dead-code
archiving and pruning needs migration.

keynslug:
> drop the old cli?

You mean, also in this PR?

zmstone:
> > drop the old cli?
> 
> You mean, also in this PR?

yeah, why not.

src/hocon_cli.erl
```diff
@@ -1,5 +1,5 @@
 %%--------------------------------------------------------------------
-%% Copyright (c) 2021-2023 EMQ Technologies Co., Ltd. All Rights Reserved.
+%% Copyright (c) 2026 EMQ Technologies Co., Ltd. All Rights Reserved.
```
zmstone:
```suggestion
%% Copyright (c) 2021-2026 EMQ Technologies Co., Ltd. All Rights Reserved.
```

src/hocon_cli.erl
```diff
@@ -129,592 +364,392 @@ main(Args) ->
...
+    %% NOTE: Adapted from legacy CLI.
+    {FmtArg, FmtValue} =
+        case Value of
+            V when is_binary(V) -> {"~s", V};
+            #{} -> {"~s", <<"{...}">>};
+            [] -> {"~s", <<"[]">>};
+            [_ | _] -> {"~s", <<"[...]">>};
+            V -> {"~0tp", V}
+        end,
+    logger:log(Level, "~ts [~ts]: " ++ FmtArg, [Var, Path, FmtValue]);
```
keynslug: Q @zmstone: currently, this will cause env overrides to be emitted on `info` level to stderr, example below. Do we need to preserve the old behavior where those were printed to stdout?

```
2026-09-07T15:45:04.640266+02:00 [info] EMQX_FOO__MAX [foo.max]: 42
```

zmstone: yes, this is because when such lines are printed, EMQX log handlers are not installed yet.

zmstone (commented):
should keep the file archive and prune?

keynslug:
> should keep the file archive and prune?

Working on it on EMQX side

keynslug:
> should keep the file archive and prune?

https://github.com/emqx/emqx/pull/18834

src/hocon_cli.erl
```diff
@@ -129,592 +364,392 @@ main(Args) ->
...
-            prepare_staged_files(StagedFiles, [PreparedFile | PreparedFiles]);
-        {error, Reason} ->
-            {error, Filename, Reason, PreparedFiles}
+write_generated_config(Args, Generated) ->
+    AppConfig = proplists:delete(vm_args, Generated),
+    AppOutput = maps:get(config_output, Args),
+    VMArgsOutput = maps:get(vm_args_output, Args),
+    AppContent = io_lib:fwrite("~p.~n", [AppConfig]),
+    VMArgs = stringify(proplists:get_value(vm_args, Generated)),
+    case write_output(AppOutput, AppContent) of
```
zmstone, line 655: These two outputs are still written sequentially without a collision check or rollback. If both options name the same path, the command returns success but the VM-args write overwrites the app config; if the second write fails, it returns `STATUS_OUTPUT_ERROR` while leaving the new app config behind. Please reject colliding destinations and commit the pair atomically, or roll back the first write.

src/hocon_cli.erl
```diff
@@ -129,592 +364,392 @@ main(Args) ->
...
-            Conf
+get_values(#{keys := Keys} = Parsed) ->
+    with_schema_and_conf(
+        Parsed,
+        fun(Schema, Conf) ->
+            try
+                RootNames = lists:usort([root_name(Schema, Key) || Key <- Keys]),
+                %% Validate converted values, then preserve their HOCON representation for output.
+                {_, _CheckedConf} = hocon_tconf:map(Schema, Conf, RootNames, tconf_opts()),
+                {_, SerializableConf} = hocon_tconf:map(Schema, Conf, RootNames, tconf_opts_get()),
```
zmstone: The serializable pass needs to consume `CheckedConf`, not the original `Conf`. `make_serializable` mode assumes converter output as its input; with a reversible converter such as `one -> 1` and `1 -> one` in serializable mode, this code runs the forward conversion again and `get` prints `1` instead of `one`. Please feed the result of the validating pass into this call.

zmstone: this is new ?

is there anything wrong printing the checked value ?

keynslug: This is new in the sense that non-scalar values are printed as inline HOCON.

In the original CLI, they are printed as `~0tp`, which I found to be odd and decided to change: the idea that HOCON CLI should give users outputs in well-known formats. `_CheckedConf` may contain converted values, testcase in 06d389852cff516734eda5e19ba7690a48aa2abb illustrates that.

zmstone: recommend to restore the old print.
main concern is the time cost for little gain
the map function walks the whole config tree again for serializable

keynslug: The gain is clean CLI UX in this case. New CLI does not worth the effort if it's going to inherit all the oddities from the old one IMO.

>  the map function walks the whole config tree again for serializable

I think this can be solved by other means, I'll see what I can do.

keynslug: Took a stab at it in ef08d4626773019076d924185b93c0942d7ed406.

src/hocon_maps.erl
```diff
@@ -158,6 +158,43 @@ get(Path, Map) ->
             do_get(hocon_util:split_path(Path), Map, map)
     end.
 
+%% @doc Get the source value from a maybe-rich map.
+%% Falls back to the current value when no source value was preserved.
+-spec get_source([name()] | name(), config()) -> term().
+get_source(Path, Map) ->
+    case is_richmap(Map) of
+        true ->
+            source_to_plain(deep_get(Path, Map));
```
zmstone, line 167: [P2] Do not traverse checked strings as arrays

`deep_get/2` unwraps every rich-map node and then treats any Erlang list as an indexable array. Checked `string()` values are Erlang charlists, so this makes numeric suffixes address characters. With the bundled demo schema/config, `hocon get ... foo.setting.1` prints `104` for `"hello"`; the same path is not a real configuration key. Please retain/consult `$hcTyp` while walking the rich map so indexing is only allowed for array nodes (and add this scalar-path case beside the out-of-range array tests).

src/hocon_tconf.erl
```diff
@@ -1109,12 +1123,60 @@ unbox(Boxed) ->
...
+    ValueIn,
+    ValuePlain,
+    #{?HOCON_V := _} = Box
+) when ValueIn =/= undefined ->
+    with_source(ValuePlain, Box);
+preserve_source(_Opts, _ValueIn, _ValuePlain, Box) ->
+    Box.
+
+preserve_converted_source(Opts, Type, ValueIn, ValueOut) ->
+    case is_primitive_type(Type) of
```
zmstone: [P2] Preserve input across converters on primitive unions

This guard classifies every union as non-primitive, so a field-level converter on a union loses its original HOCON value before the selected primitive member is mapped. For example, a `union([a, b])` field whose converter maps `<<"old">>` to `a` produces `get_source(...) = a`, not `<<"old">>`, even with `keep_source => true`. This makes `get` expose the converted value again and can reintroduce the non-HOCON output problem this mode is meant to avoid when a converter feeds a custom primitive union member. Please preserve the pre-converter source when the selected union branch is primitive (or propagate it into that branch).

zmstone (commented):
looks good, but codex has some comments.

src/hocon_maps.erl
keynslug: I believe this is actually pre-existing issue, manifesting also in `hocon_pp:flat_dump/1` for example. Tried to address issues of this class in recent commits.

zmstone (dismissed):

keynslug:
Sorry, found few more edge cases related to `source` mapping and array-vs-string ambiguity.

zmstone (approved):

